PR #2, opened 2020-02-03
composer.json adapted so that you can install the plugin via composer.

mrcodefinger:
With this modification of composer.json we could provide the plugin to [https://packagist.org](https://packagist.org) and are able to deploy the plugin via composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,12 @@
 		"phpstan/phpstan": ">=0.9",
 		"php-coveralls/php-coveralls": "^2.1"
 	},
+	"repositories": [
+		{
+			"type": "vcs",
+			"url": "https://github.com/onOfficeGmbH/sdk"
+		}
+	],
 	"autoload": {
 		 "psr-4": {
 			"onOffice\\SDK\\": "SDK/SDK/",
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,28 @@
 {
+	"name": "onofficegmbh/oo-wp-plugin",
+	"type": "wordpress-plugin",
+	"license": "GNU AGPLv3 and GNU GPLv3",
+	"description": "onOffice plugin for WordPress",
+	"homepage": "https://github.com/onOfficeGmbH/oo-wp-plugin",
+	"authors": [
+		{
+			"name": "Jakob Jungmann",
+			"email": "j.jungman@onoffice.com",
+			"homepage": "https://github.com/jayay"
+		},
+		{
+			"name": "Ana Ivanova",
+			"email": "a.ivanova@onoffice.com",
+			"homepage": "https://github.com/anaivanova"
+		}
+	],
+	"keywords": [
+		"wordpress",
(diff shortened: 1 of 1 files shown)


mrcodefinger:
I added the link to the sdk repository so you can install the submodule with the `--prefer-source` parameter as well.

jayay:
Hi! Thank you for the contribution! 🤝 

I really like Composer. But is it suitable for WP plugins? Could you maybe help me out and explain what exactly the point of having a WP plugin installed with Composer is? We appreciate Composer a lot to manage our dependencies, I just think there are easier ways to install and keep plugins up-to-date. I do use [WP CLI](https://wp-cli.org/) to manage my plugins and trying out new (not yet released) features works by switching the branch on a git clone. Also, our plugin is available on [wpackagist.org](https://wpackagist.org/search?q=onoffice&type=any&search=) which obtains its packages from the official WP Plugin directory and causes us absolutely no additional work.

I'm not generally against the Idea of adding our plugin to packagist, I'm just asking myself if it is necessary. I would rather keep focused on having everything done right on the official plugin directory.

BTW: we've published our [SDK on packagist.org](https://packagist.org/packages/onoffice/sdk), so getting rid of git submodules may be a good next step once the package has a proper autoloader.

Best regards!

mrcodefinger:
Hi, you're welcome, thank you for your message. Composer is suitable for plugins if you manage Wordpress itself with composer. The most popular examples are probably [bedrock ](https://roots.io/bedrock/) or the one from [johnpbloch](https://github.com/johnpbloch/wordpress). Otherwise it is also very easy to manage Wordpress with composer and this way it has a central place to manage all dependencies including plugins and themes.
Great to read that the SDK now also has a composer.json, then I will adapt the composer.json here as well.
Kind regards!

mrcodefinger:
In addition, it should be mentioned that it is much easier to combine Wordpress specific libraries, components etc. with non-Wordpress specific ones.

jayay:
Hi! Setting up complete instances of WordPress just by using configuration sounds very tempting, but neither of the aforementioned frameworks in this PR are official. In addition to that, the commit introduces another dependency which is not trivial since we need to ship them together with our code.

I'd like to close this PR since we'd like to develop this plugin by using the official WordPress workflow and we cannot maintain "the composer way" next to that because of the additional maintenance work.

I hope this clears everything up, if you have more ideas on how to simplify things, feel free to open an issue and start a discussion.

mrcodefinger:
Hi! Too bad, I don't see any additional effort except to extend the composer.json, which I would have done with this PR. What other dependencies have I introduced?  Anyway, I can't seem to change it anyway.
Kind regards!

jayay:
Don't hesitate to ask!
We'd have to make sure installing the plugin with composer is still a working option _after every release_. This cannot be done right now, since this is not the workflow of anyone of us.

The new dependency is this one: `"composer/installers": "~1.0"`

Have a nice weekend!

PR #15, opened 2020-07-24
Add wp_link_pages function

FD-Websolutions:
The defaultviews had no pagination, so I added the wp_link_pages function.

### DIFF
--- a/templates.dist/estate/default.php
+++ b/templates.dist/estate/default.php
@@ -99,6 +99,14 @@ $dontEcho = array("objekttitel", "objektbeschreibung", "lage", "ausstatt_beschr"
 			</div>
 		</div>
 	<?php endwhile; ?>
+	<div class="oo-pagination">
+		<?php wp_link_pages(
+			array(
+				'link_before'      => '<span>',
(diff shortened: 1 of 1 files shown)


jayay:
Sadly, the current page number is a shared (global) state in WP, which, in my opinion, makes the template responsible for the pagination. I'm using the vanilla Twentytwenty theme together with the vanilla _default.php_ estate template and it works out of the box (see image below).
image

On the other hand, this also adds pagination to the detail view, which I'm pretty sure is a bug. Is there any case in which page numbers are needed in the detail view?

FD-Websolutions:
The vanilla Twentytwenty theme is not representative for productive WordPress websites. 

In most (nearly all) cases we work on, the pagination is not part of the theme, so our customers ask us for support, which takes time we need to work on our own projects.

When you add the pagination to the default.php, the default-detail.php is not affected, so I do not understand your last question.

mrcodefinger:
What's the status on this topic? Ticket #1921025

jayay:
This is going to be a project, closing this for now.

PR #17, opened 2020-09-02
Remove IDE data from repository

mrcodefinger:
Hi!
I have added the .idea folder to the .gitignore. On the one hand my IDE settings are partly overwritten, on the other hand I think that this data does not belong in a repository. I recommend to add the .idea folder to the .gitignore_global.
Kind regards!


jayay:
This was done on purpose as it is the only way for us to keep our settings in sync. It is also an [officially documented](https://www.jetbrains.com/help/phpstorm/creating-and-managing-projects.html#share-project-through-vcs) [feature of Intellij IDEs](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839). So why not?

mrcodefinger:
Each development environment has its own optimizations and sometimes different platforms, tasks, settings and folder structures. Every developer should have the freedom to use his own IDE in his own productive way. IDE's are local tools, they are not linked to the project. If not, .vscode, .elipse, .vim and so on should also be considered. I will not blindly trust the IDE creators, for them it is a business, for me it is just a tool. If I'm informed correctly, you should share global settings via the .editorconf in a separate repository if you want to, but this overrides the built-in cloudsync function of PHPStorm.

jayay:
Please see here: https://www.jetbrains.com/help/idea/sharing-your-ide-settings.html

> For information on how to share settings related to specific projects, see Share project settings through VCS.

The project specific settings are _supposed to be shared through VCS_.

Each developer can configure their IDE in their own way. But project specific settings require the biggest amount of time and it is only a pleasure to each individual developer if they come just with the project. We use this feature internally as well and from my experience I can tell it is so much harder if each individual developer needs to configure the exact same settings in order to get the same setup. I would be open to fixing the issue that you have encountered while getting this running inside your IDE, but deleting the settings entirely would be counter-productive.

mrcodefinger:
In the documentation you posted there it is also recommended to share settings via a separate repository ;-)
[https://www.jetbrains.com/help/idea/sharing-your-ide-settings.html#import-export-settings](https://www.jetbrains.com/help/idea/sharing-your-ide-settings.html#import-export-settings)

pmaasz:
It is uncommon to publish the .idea/ directory in the same repo as the code.

jayay:
I asked around and no one seemed to need the `.idea` directory. Will merge once the master branch is clean.

PR #24, opened 2020-11-05
CSS files concatenated to save requests...

mrcodefinger:
...to improve page speed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,14 @@ composer-install-nodev:
 pot:
 	vendor/bin/wp i18n make-pot . languages/onoffice.pot --skip-js
 	sed -i -r "s/onOffice for WP-Websites \(dev\)/onOffice for WP-Websites/" languages/onoffice.pot
+
+concat-css:
+	@cat `ls -I ./third_party/slick/*\.css`  ./css/*\.css > ./css/build/oo-wp-plugin.css
+
+mkdirs:
+	@mkdir ./css/build/
 	
-release: pot copy-files-release change-title add-version composer-install-nodev
+release: pot copy-files-release change-title add-version composer-install-nodev mkdirs concat-css
 
 test-zip: pot copy-files-release add-version composer-install-nodev
 	cd $(PREFIX); zip -r onoffice.zip onoffice/
--- a/plugin/ScriptLoader/ScriptLoaderGenericConfigurationDefault.php
+++ b/plugin/ScriptLoader/ScriptLoaderGenericConfigurationDefault.php
@@ -67,12 +67,7 @@ class ScriptLoaderGenericConfigurationDefault
 			(new IncludeFileModel($script, 'onoffice-form-preview', plugins_url('/js/onoffice-form-preview.js', $pluginPath)))
 				->setLoadInFooter(true),
 
-			new IncludeFileModel($style, 'onoffice-default', plugins_url('/css/onoffice-default.css', $pluginPath)),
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -24,8 +24,14 @@ composer-install-nodev:
 pot:
 	vendor/bin/wp i18n make-pot . languages/onoffice.pot --skip-js
 	sed -i -r "s/onOffice for WP-Websites \(dev\)/onOffice for WP-Websites/" languages/onoffice.pot
+
+concat-css:
+	@cat `ls -I ./third_party/slick/*\.css`  ./css/*\.css > ./css/build/oo-wp-plugin.css
+
+mkdirs:
+	@mkdir ./css/build/
```
jayay, line 32: What do you think about not having the `build` sub-directory, but naming the new CSS file something like `all.css`?

jayay: the `mkdirs` target could be dropped then. I would really like that.

plugin/ScriptLoader/ScriptLoaderGenericConfigurationDefault.php
```diff
@@ -67,12 +67,7 @@ public function getScriptLoaderGenericConfiguration(): array
 			(new IncludeFileModel($script, 'onoffice-form-preview', plugins_url('/js/onoffice-form-preview.js', $pluginPath)))
 				->setLoadInFooter(true),
 
-			new IncludeFileModel($style, 'onoffice-default', plugins_url('/css/onoffice-default.css', $pluginPath)),
```
jayay, line 70: Deleting the old includes would make life really hard during development. What about this:
If the concatenated CSS file exists, load that, if it doesn't, load all the others.

jayay (changes requested):
I've annotated some lines. Other than those, what's the deal with the empty `*.css` file. Was that on purpose?

Makefile
mrcodefinger: > What do you think about not having the `build` sub-directory, but naming the new CSS file something like `all.css`?
I have nothing against all.css, but we can't do without a subdirectory, because `./css/*\.css` searches for all CSS files in this folder and in another process all.css would be in there as well.

plugin/ScriptLoader/ScriptLoaderGenericConfigurationDefault.php
mrcodefinger: Yes, good idea!

Makefile
mrcodefinger: And in another pull request I will summarize the JavaScript files, I think we will not be able to avoid mkdirs.

mrcodefinger:
> I've annotated some lines. Other than those, what's the deal with the empty `*.css` file. Was that on purpose?

`*.css` is a wildcard for all CSS files in this directory.

Makefile
jayay: Dealing with the Javascript in a second PR sounds good. I'm gonna need a little bit of time to think through the mkdir, though.

jayay:
I guess that means the literal [*.css](https://github.com/onOfficeGmbH/oo-wp-plugin/blob/559391a7489c9c5d4d7f9ae32bb1ba2dca5955e8/css/*.css) file can be removed then 👍

Makefile
mrcodefinger: Maybe I should rename 'build' into 'dist'?

jayay:
The idea was good, but I'm closing this due to inactivity. It is better for onOffice employees to create a ticket that can in turn be the template for a project. So feel free to create a ticket if you think this is necessary.

PR #29, opened 2020-11-24
P#39191 similar estates

nglelinh:
P#39191 Fields for 'Ähnliche Immobilien' should be configurable (18656)


plugin/Controller/EstateViewSimilarEstates.php
```diff
@@ -96,6 +97,8 @@ private function buildEstateParameters(EstateListBase $pEstateList): Generator
 			$this->_pFilterConfiguration->setStreet($pValuesCurrentEstate->getValueRaw('strasse') ?? '');
 			$this->_pFilterConfiguration->setCountry($pValuesCurrentEstate->getValueRaw('land') ?? '');
 			$this->_pFilterConfiguration->setPostalCode($pValuesCurrentEstate->getValueRaw('plz') ?? '');
+            $this->_pFilterConfiguration->setShowArchive($pValuesCurrentEstate->getValueRaw('status2') ?? '');
```
jayay, line 100: Mixed tabs and spaces

plugin/Model/FormModelBuilder/FormModelBuilderSimilarEstateSettings.php
```diff
@@ -0,0 +1,374 @@
...
+     *
+     * @return InputModelDB
+     *
+     */
+
+    public function createInputModelDontShowArchived()
+    {
+        $pDataViewSimilarEstates = $this->_pDataSimilarView->getDataViewSimilarEstates();
+
+        $labelDontShowArchived= __('Don&#8217;t show archived estates', 'onoffice');
```
jayay, line 272: Text within `__()` should be plain and not HTML entity encoded. Encoding and escaping should happing where output starts.

plugin/ViewFieldModifier/EstateViewFieldModifierTypeDetailSimilarEstates.php
```diff
@@ -36,6 +36,8 @@ class EstateViewFieldModifierTypeDetailSimilarEstates
 		'laengengrad',
 		'breitengrad',
 		'plz',
+        'status2',
```
jayay, line 39: `status` should be used

tests/TestClassDataViewSimilarEstates.php
```diff
@@ -48,6 +48,8 @@ public function testDefaultValues()
 		$this->assertEquals(true, $pDataViewSimilarEstates->getSameEstateKind());
 		$this->assertEquals(true, $pDataViewSimilarEstates->getSameMarketingMethod());
 		$this->assertEquals(false, $pDataViewSimilarEstates->getSamePostalCode());
+        $this->assertEquals(false, $pDataViewSimilarEstates->getDontShowArchived());
```
jayay, line 51: Indents are funny here

tests/TestClassFilterConfigurationSimilarEstates.php
```diff
@@ -56,6 +56,8 @@ public function setUp()
 		$this->_pDataViewSimilarEstates->setSameEstateKind(true);
 		$this->_pDataViewSimilarEstates->setSameMarketingMethod(true);
 		$this->_pDataViewSimilarEstates->setSamePostalCode(true);
+        $this->_pDataViewSimilarEstates->setDontShowArchived(true);
```
jayay, line 59: These conditions are not being checked in the test

tests/TestClassEstateViewFieldModifierTypeDetailSimilarEstates.php
```diff
@@ -56,6 +56,8 @@ class TestClassEstateViewFieldModifierTypeDetailSimilarEstates
 		'laengengrad',
 		'breitengrad',
 		'plz',
+        'referenz',
```
jayay, line 59: indent is off

tests/TestTemplateEstateDefaultDetail.php
```diff
@@ -52,7 +52,7 @@ public function prepare()
 				'getEstatePictureTitle',
 				'getDocument',
 				'getCurrentEstateId',
-				'getSimilarEstates',
+                'getSimilarEstates',
```
jayay, line 55: this file can be reverted

jayay (changes requested):
Thank you for the hard work! UI in our WP plugin is a challenge, but you did very well 👏🏻 

The overall result looks very promising, but I am concerned about backwards compability. Did you happen to test if a migration works correctly? I didn't try it out, but existing websites using this plugin must not break after an update. The `DatabaseChanges` class might be the place where the upgrade logic will have to go.

Other than that, there's load of classes with inconsistent indents, and a [decrease in code coverage](https://coveralls.io/github/onOfficeGmbH/oo-wp-plugin?branch=7905-similar-estates) by 0.4% compared to master. Hints about missing coverage can be found in one of the tabs on that page. I'd appreciate if some more lines could be covered.

Thank you!

plugin/Filter/DefaultFilterBuilderSimilarEstates.php
```diff
@@ -84,17 +84,18 @@ public function buildFilter(): array
 			$filter['plz'] []= ['op' => '=', 'val' => $pFilterConfiguration->getPostalCode()];
 		}
 
-        if ($pDataListView->getDontShowArchived() && $showArchived !== 'status2obj_archiviert') {
-            $filter['status2'] []= ['op' => '!=', 'val' => ['status2obj_archiviert'] ];
-        }
+		if ($pDataListView->getShowArchived() && $showArchived !== 'status2obj_archiviert') {
+			$filter['status2'] [] = ['op' => '!=', 'val' => ['status2obj_archiviert']];
```
jayay, line 88: The getter got inverted, but the boolean logic and the filter weren't. I guess that's not on purpose?

`status2` is still an issue.

plugin/Filter/DefaultFilterBuilderSimilarEstates.php
```diff
@@ -84,17 +84,18 @@ public function buildFilter(): array
...
+		}
 
-        if ($pDataListView->getDontShowReference() && $showReference !== '1') {
-            $filter['referenz'] []= ['op' => '!=', 'val' => '1' ];
-        }
+		if ($pDataListView->getShowReference() && $showReference !== '1') {
```
jayay, line 91: Same as above

plugin/Installer/DatabaseChanges.php
```diff
@@ -138,6 +138,11 @@ public function install()
 			$dbversion = 16;
 		}
 
+		if ($dbversion == 16) {
+			dbDelta( $this->migrationsDataSimilarEstates() );
```
jayay, line 142: The `dbDelta` WP function is not helpful here, because it takes a `CREATE TABLE` statement and compares it to the currently existing schema and modifies it into an `ALTER TABLE`. What's being done here is to just insert a row.
The `dbDelta` function call can just be removed and `$this->migrationsDataSimilarEstates()` should be called directly.

plugin/Installer/DatabaseChanges.php
```diff
@@ -482,6 +487,47 @@ private function updateSortByUserDefinedDefault()
...
 	}
 
+	/**
+	 *
+	 */
+
+	public function migrationsDataSimilarEstates()
+	{
+		$prefix = $this->getPrefix();
+		$defaultViewOptions = $this->_pWPDB->get_results("select * from " . $prefix . "options where option_name = 'onoffice-default-view'");
```
jayay, line 497: `get_option()` can be used here.

plugin/Installer/DatabaseChanges.php
```diff
@@ -482,6 +487,47 @@ private function updateSortByUserDefinedDefault()
...
+		if(!empty($defaultViewOptions)){
+			add_option('onoffice-similar-estates-settings-view', '');
+			$old_data = [];
+
+			foreach ($defaultViewOptions as $defaultViewOption) {
```
jayay, line 502: There should be only one row with the detail view (it's a serialized `onOffice\WPlugin\DataView\DataDetailView`). How does this work?

plugin/Model/FormModelBuilder/FormModelBuilderSimilarEstateSettings.php
```diff
@@ -259,62 +281,59 @@ public function createInputModelSimilarEstateMarketingMethod()
...
 
-    public function createInputModelDontShowArchived()
-    {
-        $pDataViewSimilarEstates = $this->_pDataSimilarView->getDataViewSimilarEstates();
+	public function createInputModelShowArchived()
+	{
+		$pDataViewSimilarEstates = $this->_pDataSimilarView->getDataViewSimilarEstates();
 
-        $labelDontShowArchived= __('Don&#8217;t show archived estates', 'onoffice');
+		$labelShowArchived = __('Don\'t show archived estates', 'onoffice-for-wp-websites');
```
jayay, line 293: This is inverted as well

plugin/Model/FormModelBuilder/FormModelBuilderSimilarEstateSettings.php
```diff
@@ -259,62 +281,59 @@ public function createInputModelSimilarEstateMarketingMethod()
...
 
-    public function createInputModelDontShowReference()
-    {
-        $pDataViewSimilarEstates = $this->_pDataSimilarView->getDataViewSimilarEstates();
+	public function createInputModelShowReference()
+	{
+		$pDataViewSimilarEstates = $this->_pDataSimilarView->getDataViewSimilarEstates();
 
-        $labelDontShowReference= __('Don&#8217;t show reference estates', 'onoffice');
+		$labelShowReference = __('Don\'t show reference estates', 'onoffice-for-wp-websites');
```
jayay, line 314: This is inverted too

plugin/ViewFieldModifier/EstateViewFieldModifierTypeDetailSimilarEstates.php
```diff
@@ -36,8 +36,8 @@ class EstateViewFieldModifierTypeDetailSimilarEstates
 		'laengengrad',
 		'breitengrad',
 		'plz',
-        'status2',
-        'referenz',
+		'status2',
```
jayay, line 39: `s/status2/status/g` 😉

tests/TestClassDataDetailView.php
```diff
@@ -69,6 +69,38 @@ class TestClassDataDetailView
...
+	/**
+	 *
+	 */
+
+	public function test__wakeup()
+	{
+		$pInstance = new DataDetailView();
+
+		// old versions didn't have $_pDataViewSimilarEstates, but it must always be set
+		$pClosure = function() {
```
jayay, line 84: A way to solve this without `Closure::bind()` would be much nicer, for example by using a fixed string of that serialized object.

tests/TestClassDataSimilarView.php
```diff
@@ -33,31 +33,31 @@ class TestClassDataSimilarView
...
+		'objekttitel',
+		'objektnr_extern',
+		'regionaler_zusatz',
+		'kaufpreis',
+		'wohnflaeche',
+		'anzahl_zimmer',
+		'kaltmiete',
+		'ort',
+		'plz',
+		'status2',
```
jayay, line 46: 🚨 `status2` 🚨

tests/TestClassDataViewSimilarEstates.php
```diff
@@ -48,8 +48,8 @@ public function testDefaultValues()
 		$this->assertEquals(true, $pDataViewSimilarEstates->getSameEstateKind());
 		$this->assertEquals(true, $pDataViewSimilarEstates->getSameMarketingMethod());
 		$this->assertEquals(false, $pDataViewSimilarEstates->getSamePostalCode());
-        $this->assertEquals(false, $pDataViewSimilarEstates->getDontShowArchived());
-        $this->assertEquals(false, $pDataViewSimilarEstates->getDontShowReference());
+		$this->assertEquals(false, $pDataViewSimilarEstates->getShowArchived());
```
jayay, line 51: is `false` still correct as the default value? (in both cases)

tests/TestClassDataViewSimilarEstates.php
```diff
@@ -102,7 +102,7 @@ public function testOverriddenFields()
 			'kaltmiete',
 			'ort',
 			'plz',
-            'status2',
+			'status2',
```
jayay, line 105: Another `status2`

tests/TestClassDatabaseChanges.php
```diff
@@ -70,6 +70,21 @@ public function prepare()
 
 		$this->_pWpOption = new WPOptionWrapperTest();
 		$this->_pDbChanges = new DatabaseChanges($this->_pWpOption, $wpdb);
+
+		add_option('onoffice-default-view', '');
```
jayay, line 74: instead of using `add_option()` and `update_option()`, `$this->_pWpOption` can be used. That's especially handy for writing tests because it can be mocked.

tests/TestClassDatabaseChanges.php
```diff
@@ -84,7 +99,42 @@ public function testInstall(): array
...
+	public function testInstallMigrationsDataSimilarEstates(): array
+	{
+		add_option('oo_plugin_db_version', '16');
+		add_filter('query', [$this, 'saveCreateQuery'], 1);
+		$this->_pDbChanges->install();
+		remove_filter('query', [$this, 'saveCreateQuery'], 1);
+
+		$prefix = $this->_pDbChanges->getPWPDB()->prefix;
+		$similarViewOptions = $this->_pDbChanges->getPWPDB()->get_results("select * from " . $prefix . "options where option_name = 'onoffice-similar-estates-settings-view'");
+		$new_data = [];
```
jayay, line 115: See comment in the implementation class

tests/TestClassDefaultFilterBuilderSimilarEstates.php
```diff
@@ -63,8 +63,8 @@ public function setUp()
 		$this->_pFilterConfigurationSimilarEstates->setEstateKind('wohnung');
 		$this->_pFilterConfigurationSimilarEstates->setMarketingMethod('kauf');
 		$this->_pFilterConfigurationSimilarEstates->setPostalCode('52070');
-        $this->_pFilterConfigurationSimilarEstates->setShowArchive('status2obj_archiviert');
-        $this->_pFilterConfigurationSimilarEstates->setShowReference('1');
+		$this->_pFilterConfigurationSimilarEstates->setShowArchive('status2obj_aktiv');
```
jayay, line 66: Isn't the parameter supposed to be of type `bool`?

jayay: In both cases

tests/TestClassDefaultFilterBuilderSimilarEstates.php
```diff
@@ -144,48 +144,47 @@ public function testBuildFilterSamePostalCode()
...
+	public function testBuildFilterShowArchive()
+	{
+		$this->getDataViewSimilarEstates()->setShowArchived(true);
+		$result = $this->_pDefaultFilterBuilderSimilarEstates->buildFilter();
+		$expectation = [
+			'veroeffentlichen' => [
+				['op' => '=', 'val' => 1],
+			],
+			'status2' => [
+				['op' => '!=', 'val' => ['status2obj_archiviert']],
```
jayay, line 160: This wasn't inverted either

tests/TestClassDefaultFilterBuilderSimilarEstates.php
```diff
@@ -144,48 +144,47 @@ public function testBuildFilterSamePostalCode()
...
+	public function testBuildFilterShowReference()
+	{
+		$this->getDataViewSimilarEstates()->setShowReference(true);
+		$result = $this->_pDefaultFilterBuilderSimilarEstates->buildFilter();
+		$expectation = [
+			'veroeffentlichen' => [
+				['op' => '=', 'val' => 1],
+			],
+			'referenz' => [
+				['op' => '!=', 'val' => '1'],
```
jayay, line 180: This condition wasn't inverted

tests/TestClassEstateViewFieldModifierTypeDetailSimilarEstates.php
```diff
@@ -56,8 +56,8 @@ class TestClassEstateViewFieldModifierTypeDetailSimilarEstates
 		'laengengrad',
 		'breitengrad',
 		'plz',
-        'referenz',
-        'status2',
+		'referenz',
+		'status2',
```
jayay, line 60: `status2` should be replaced

tests/TestClassFormModelBuilderSimilarEstateSettings.php
```diff
@@ -37,61 +37,228 @@ class TestClassFormModelBuilderSimilarEstateSettings
...
+
+	/** @var DataSimilarView */
+	private $_pDataSimilarView = null;
+
+	/**
+	 * @before
+	 */
+	public function prepare()
+	{
+		$this->_pInputModelSimilarViewFactory = new InputModelOptionFactorySimilarView('onoffice');
```
jayay, line 71: `Factory` in the attribute name is confusing

jayay (changes requested):
Thank you for your work. It's gotten better. There are a few major things to be resolved:

- The migration into the new Similar Estates View needs to be fixed. See my comments in the `DatabaseChanges` class.
- There's some confusion about the former `dont` attributes: those were renamed but the logic around them wasn't inverted.
- `status2` is still around. I suggest you to change this at the very end because we might need some clever ideas here.

tests/TestClassFormModelBuilderSimilarEstateSettings.php
jayay: That one was a mistake, sorry.

plugin/Installer/DatabaseChanges.php
```diff
@@ -482,6 +487,42 @@ private function updateSortByUserDefinedDefault()
...
+			$dataDetailViewSimilarEstates = $defaultViewOptions->getDataViewSimilarEstates();
+			$old_data['radius'] = $dataDetailViewSimilarEstates->getRadius();
+			$old_data['same_kind'] = $dataDetailViewSimilarEstates->getSameEstateKind();
+			$old_data['same_maketing_method'] = $dataDetailViewSimilarEstates->getSameMarketingMethod();
+			$old_data['same_postal_code'] = $dataDetailViewSimilarEstates->getSamePostalCode();
+			$old_data['amount'] = $dataDetailViewSimilarEstates->getRecordsPerPage();
+			$old_data['similar_estates_template'] = $dataDetailViewSimilarEstates->getTemplate();
+
+			$dataSimilarViewOptions = new \onOffice\WPlugin\DataView\DataSimilarView();
+			$dataSimilarViewOptions->name = "onoffice-similar-estates-settings-view";
```
jayay, line 511: Why is this undeclared property `name` being set?

plugin/Installer/DatabaseChanges.php
```diff
@@ -482,6 +487,42 @@ private function updateSortByUserDefinedDefault()
...
+			$old_data = [];
+			$old_data['enablesimilarestates'] = $defaultViewOptions->getDataDetailViewActive();
+
+			$dataDetailViewSimilarEstates = $defaultViewOptions->getDataViewSimilarEstates();
+			$old_data['radius'] = $dataDetailViewSimilarEstates->getRadius();
+			$old_data['same_kind'] = $dataDetailViewSimilarEstates->getSameEstateKind();
+			$old_data['same_maketing_method'] = $dataDetailViewSimilarEstates->getSameMarketingMethod();
+			$old_data['same_postal_code'] = $dataDetailViewSimilarEstates->getSamePostalCode();
+			$old_data['amount'] = $dataDetailViewSimilarEstates->getRecordsPerPage();
+			$old_data['similar_estates_template'] = $dataDetailViewSimilarEstates->getTemplate();
```
jayay, line 508: The `fields` should be copied from the old view into the new one too.

plugin/DataView/DataDetailView.php
```diff
@@ -97,6 +100,22 @@ class DataDetailView
 	/** @var int */
 	private $_movieLinks = MovieLinkTypes::MOVIE_LINKS_NONE;
 
+	/** @var bool */
+	private $_dataDetailViewActive = false;
+
+	/** @var DataViewSimilarEstates */
+	private $_pDataViewSimilarEstates = null;
```
jayay: A comment would be good that states that the property `private $_pDataViewSimilarEstates` and its getter need to stay for compatibility reasons.

plugin/DataView/DataDetailView.php
```diff
@@ -117,19 +116,6 @@ public function __construct()
 		$this->_pDataViewSimilarEstates = new DataViewSimilarEstates();
 	}
 
-
-	/**
-	 *
-	 */
-
-	public function __wakeup()
```
jayay, line 125: I think this should be pulled back in since `DataViewSimilarEstates` is needed for the migration from database version 16 to 17.

jayay (changes requested):
Thank you. I've tested this and it looked good to me.

We do now have `DataViewSimilarEstates` and `DataSimilarView`. Both have the same job but are nested into each other. Is there a way to clean this up? `DataViewSimilarEstates` must be edited with caution because modifying it could break the migration step.

Could you please also see the unresolved comments and fix/answer them?

Another minor thing I've noticed is, when you need to rebuild the `onoffice-for-wp-websites.pot` file, you should use our `make pot` command, instead of calling `wp cli` directly.

plugin/Installer/DatabaseChanges.php
jayay: Also, variables for each individual value would be nicer than the current array `$old_data`.

jayay (approved):
Thanks! I'll forward this to QA for formal testing.

jayay:
@hungnc89 We tried to merge this into masters but a bunch of tests failed. The new base filters need to be added to your new methods in `\onOffice\tests\TestClassDefaultFilterBuilderSimilarEstates`, but `onOffice\tests\TestClassDatabaseChanges::testInstall` and `onOffice\tests\TestClassEstateList::testGetEstateLink` both fail with 
```
WordPress database error Table 'wp_test.wptests_oo_plugin_cache' doesn't exist for query
                                DELETE
                                FROM wptests_oo_plugin_cache
                                WHERE UNIX_TIMESTAMP(cache_created) < 1612433095
```

That means there's an attempt to connect to the DB through the onOffice SDK.
Would you please have a look and open a new PR?

PR #31, opened 2020-11-25
Fix version in plugin.php doc block.

mrcodefinger:


jayay (commented):
Version `2.0` is not the actual version. The version string gets written during the creation of a release, which is the name of the current git branch. See [Makefile](https://github.com/onOfficeGmbH/oo-wp-plugin/blob/a09ac949f379aa899e27eb9ce6b52e86b02b85ff/Makefile#L19).

PR #34, opened 2020-12-08
fix-bug-jquery-removeProp-update-wordpress-version5.6

hungnc89:


jayay:
This looks good to me, thank you. If there's nothing else to do on this, would you please also increase the [`tested up to:` value in our readme.txt](https://github.com/onOfficeGmbH/oo-wp-plugin/blob/a6f51c4e94817260787642cf8a879ea2ed0f11cc/readme.txt#L5)?

jayay (approved):

PR #38, opened 2020-12-21
8540 add github action

nglelinh:
@jayay 
please add the following environment value for testing the deploy action:

SVNURL: https://nglelinh@svn.code.sf.net/p/onofficewpplugin/code
SVN_USER: nglelinh
SVN_PASSWORD: XXXX (hidden)

I am using https://sourceforge.net/ for a free public SVN repository. You can also use this account to login and view the result.
PS: please change the extension of release.yml.bk to release.yml for testing the deploy action
This is my testing on another repo: https://snipboard.io/OGC4f3.jpg

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,12 @@ name: Testing
 
 # Controls when the action will run.
 on:
-  # Triggers the workflow on push or pull request events but only for the master branch
+  # Triggers the workflow on push or by cron schedule
   push:
-    branches: [ 8540-wordpress-implement-github-actions ]
+    branches:
+      -  '*'    # push to any branch
+  schedule:
+    cron: "0 13 * * 1" # Every Monday at 1PM UTC (9AM EST)
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
@@ -20,7 +23,7 @@ jobs:
       uses: mirromutth/mysql-action@v1.1
       with:
         mysql version: '5.7'
-        mysql root password: '123456'
+        mysql root password: ${{ secrets.MYSQL_PASSWORD }}
 
     - name: Checkout from GIT
       uses: actions/checkout@v2
@@ -31,7 +34,7 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: ${{ matrix.php-versions }}
-        extensions: mbstring, mysqli, intl #optional, setup extensions
+        extensions: mbstring, mysqli, intl, curl #optional, setup extensions
         coverage: xdebug #optional, setup coverage driver
 
     - name: Check PHP Version
@@ -42,7 +45,7 @@ jobs:
         composer install
 
     - name: Install WP Tests
-      run: bash bin/install-wp-tests.sh wordpress_test root '123456' 172.17.0.1 latest
+      run: bash bin/install-wp-tests.sh wordpress_test root '${{ secrets.MYSQL_PASSWORD }}' 172.17.0.1 latest
 
     - name: Run unit tests
       run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,65 @@
+name: Testing
+
+# Controls when the action will run.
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,65 @@
...
+    - name: Checkout from GIT
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php-versions }}
+        extensions: mbstring, mysqli, intl #optional, setup extensions
```
jayay, line 34: Does it make sense to mention `curl` as a required extension?

.github/workflows/main.yml
```diff
@@ -0,0 +1,65 @@
+name: Testing
+
+# Controls when the action will run.
+on:
```
jayay, line 4: Travis CI used to also test the `master` branch once a week by cron. As far as I know that's possible here as well. I thin that should be added here as well.

jayay: And every branch should be included, right?

.github/workflows/main.yml
```diff
@@ -0,0 +1,65 @@
...
+      matrix:
+        operating-system: [ubuntu-latest]
+        php-versions: ['7.0', '7.1', '7.2', '7.3']
+    name: Unittest on ${{ matrix.operating-system }} with PHP version ${{ matrix.php-versions }}
+    steps:
+    - name: Set up MySQL
+      uses: mirromutth/mysql-action@v1.1
+      with:
+        mysql version: '5.7'
+        mysql root password: '123456'
```
jayay, line 23: Is it ok to use a public password here? I do think it is, but is there any documentation on this?

.github/workflows/release.yml.bk
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    name: Deployment to WP SVN (tag only)
+    steps:
+    - name: Checkout from GIT
+      uses: actions/checkout@v2
+
+    - name: Run svn commit script
+      run: sh bin/commit-svn
```
jayay, line 20: `bin/commit-svn` requires some arguments: `$TRAVIS_TAG --stable`. Is that doable here?

jayay (changes requested):
Looks pretty good already. I'm missing a cronjob for the tests once a week. Also, is there a way to test the deployment without creating a tag? The `commit-svn` script needs the tag name as an argument.
I put your credentials into the repo settings.

.github/workflows/main.yml
nglelinh: we can transfer a secret value from env variable

.github/workflows/main.yml
nglelinh: These extensions are required for running composer install and testing. It's running without `curl`. But yes, we can add extension `curl`

nglelinh:
@jayay I fixed your comments, please help to review again
For testing, I think you can test by creating a new private repo and setup tag on that repo. Unfortunately, could not find any better way to test the tag

jayay:
The unit tests look good. In order to enable the release mechanism, I guess `.github/workflows/release.yml.bk` needs to be renamed into `.github/workflows/release.yml`?
The credentials were set up in the repository settings. The next steps to test a release would be to remove the `.travis.yml`, rename `release.yml.bk` and to create a tag, is that correct?

nglelinh:
> The unit tests look good. In order to enable the release mechanism, I guess `.github/workflows/release.yml.bk` needs to be renamed into `.github/workflows/release.yml`?
> The credentials were set up in the repository settings. The next steps to test a release would be to remove the `.travis.yml`, rename `release.yml.bk` and to create a tag, is that correct?

Yes, that's right. I didn't want to mix the 2 processes together so I changed the file extension.

jayay:
Merged in f1cb1ed989d898d6427ac1b8b97e4548a7cbc0cc. Thank you!

PR #39, opened 2020-12-24
8624 check unitest php7.4

nglelinh:


jayay (changes requested):
I guess this is the only way right now, because PHPUnit 8 (the first one to support PHP 7.4) is not supported by WordPress yet, is that correct?

Would you please consider writing `$ErrorHandler` and `$UtilPrefix` with a lower case first character and making `customErrorHandler` an anonymous function so that there will be no global function?

It would also be great if you could extend the Github workflow to run the tests automatically for PHP 7.4. 

Thank you!

jayay (approved):
Thanks!

PR #41, opened 2021-01-07
8919 select boxes with select2

hungnc89:


plugin/ScriptLoader/ScriptLoaderGenericConfigurationDefault.php
```diff
@@ -45,8 +45,11 @@ public function getScriptLoaderGenericConfiguration(): array
 		$pluginPath = ONOFFICE_PLUGIN_DIR.'/index.php';
 		$script = IncludeFileModel::TYPE_SCRIPT;
 		$style = IncludeFileModel::TYPE_STYLE;
-
 		$values = [
+			(new IncludeFileModel($script, 'select2', plugins_url('/vendor/select2/select2/dist/js/select2.min.js', $pluginPath)))
+				->setLoadInFooter(true),
+			(new IncludeFileModel($script, 'onoffice-select2', plugins_url('/js/onoffice-select2.js', $pluginPath)))
```
jayay, line 51: _onoffice-select2_ needs to have _select2_ defined as a dependency

templates.dist/fields.php
```diff
@@ -70,18 +70,19 @@ function renderFieldEstateSearch(string $inputName, array $properties) {
...
-				echo '<div data-name="'.esc_html($inputName).'[]" class="multiselect" data-values="'
-					.esc_html(json_encode($permittedValues)).'" data-selected="'
-					.esc_html(json_encode($selectedValue)).'">
-				<input type="button" class="onoffice-multiselect-edit" value="'
-					.esc_html__('Edit values', 'onoffice').'">
-			</div>
-			';
+				$htmlOptions = '';
+				foreach ($permittedValues as $key => $value) {
+					$htmlOptions .= '<option value="'.$key.'">'.$value.'</option>';
```
jayay, line 75: Escaping is missing here. 

```php
$htmlOptions .= '<option value="'.esc_attr($key).'">'.esc_html($value).'</option>';
```

jayay (changes requested):
Looks mostly good, except for the escaping part!

Did you try it out with forms that preview the amount of results before submitting?

templates.dist/fields.php
```diff
@@ -70,18 +70,19 @@ function renderFieldEstateSearch(string $inputName, array $properties) {
...
-				echo '<div data-name="'.esc_html($inputName).'[]" class="multiselect" data-values="'
-					.esc_html(json_encode($permittedValues)).'" data-selected="'
-					.esc_html(json_encode($selectedValue)).'">
-				<input type="button" class="onoffice-multiselect-edit" value="'
-					.esc_html__('Edit values', 'onoffice').'">
-			</div>
-			';
+			$htmlOptions = '';
+			foreach ($permittedValues as $key => $value) {
+				$htmlOptions .= '<option value="' . $key . '">' . $value . '</option>';
```
jayay, line 75: `esc_attr()` and `esc_html()` are missing here

js/onoffice-form-preview.js
```diff
@@ -85,6 +85,17 @@
         }
     });
 
+    document.querySelectorAll('form[data-estate-search-name]').forEach(element => {
```
jayay, line 88: HTML Forms for forms (in terms of configurable forms in the plugin settings) have a different attribute name that depends of the kind of form. This selector would only work for estate search forms.

templates.dist/fields.php
```diff
@@ -70,18 +70,19 @@ function renderFieldEstateSearch(string $inputName, array $properties) {
...
+			$htmlSelect = '<select class="custom-multiple-select form-control" name="' . esc_html($inputName) . '[]" multiple="multiple">';
+			$htmlSelect .= $htmlOptions;
+			$htmlSelect .= '</select>';
+			echo $htmlSelect;
 		} elseif ( $inputName === 'regionaler_zusatz' ) {
 			echo renderRegionalAddition($inputName, $selectedValue ?? [], true, $properties['label'], $properties['permittedvalues'] ?? null);
 		}
 		elseif ( $inputName === 'country' )	{
-			echo '<select size="1" name="'.esc_attr($inputName).'">';
+			echo '<select class="custom-single-select" size="1" name="'.esc_attr($inputName).'">';
```
jayay, line 85: Is the select2 plugin wanted for single-selects?

jayay (changes requested):
Thank you. A couple of open todo's exist:

1.  Is select2 wanted for single select inputs? That wasn't specified, so that's something open for discussion.
2. This doesn't work on forms, such as the applicant search form.
3. It is important to not only change `composer.json`, but also to commit the changes to `composer.lock` after adding the dependency. The changes to `composer.lock` can be re-added using `composer update select2/select2`.
4. Escaping appears to be missing.

The CSS could use some fine-tuning, but I would say that is a low-priority issue.

jayay (approved):
Looks good to me!

jayay (approved):

PR #43, opened 2021-01-18
Filter sold and reserved estates from similar estates.

FD-Websolutions:


plugin/Filter/DefaultFilterBuilderSimilarEstates.php
```diff
@@ -86,6 +86,9 @@ public function buildFilter(): array
 			$filter['Id'] []= ['op' => 'not in', 'val' => $this->_excludeIds];
 		}
 
+		$filter['verkauft'] []= ['op' => 'not in', 'val' => 1];
+		$filter['reserviert'] []= ['op' => 'not in', 'val' => 1];
```
jayay, line 90: Why not insert it at the top? (line 63)

jayay (changes requested):
I don't know if you've talked to Pawel about this, but it makes sense to me. Would you mind removing the `.DS_Store` files, please?

plugin/Filter/DefaultFilterBuilderSimilarEstates.php
FD-Websolutions: I changed the position.

jayay (approved):

jayay:
Released in v2.8.3.

PR #49, opened 2021-02-04
#9882 x header noindex option for pdf exposes

hungnc89:


plugin.php
```diff
@@ -163,6 +163,9 @@
 add_action('parse_request', function(WP $pWP) use ($pDI) {
 	if (isset($pWP->query_vars['document_pdf'])) {
 		try {
+			if (empty(get_option('onoffice-settings-google-bot-index-pdf-expose'))) {
+				header('X-Robots-Tag: googlebot: noindex, nofollow');
```
jayay, line 167: Would it be possible to integrate this into `PdfDownload` class, so that this behavior becomes testable?

plugin/Gui/AdminPageApiSettings.php
```diff
@@ -144,6 +145,24 @@ private function addFormModelGoogleMapsKey()
 		$this->addFormModel($pFormModel);
 	}
 
+	private function addFormModelGoogleBotSettings()
+	{
+		$labelGoogleBotIndexPdfExpose = __('Index Pdf Expose', 'onoffice-for-wp-websites');
```
jayay, line 150: The correct term for "Pdf Expose" would be `PDF brochure`. That's something that's confusing for now but should be used more, especially across the code.

jayay (changes requested):

plugin.php
hungnc89: @jayay Becomes to testable, I think `PdfDocumentFetcher` class will be matching better than `PdfDownload` class, what do you think about that?

plugin/Gui/AdminPageApiSettings.php
hungnc89: Ok I will change "Pdf Expose" to "PDF brochure"

plugin.php
```diff
@@ -169,6 +169,7 @@
 			$pPdfDocumentModel = new PdfDocumentModel($pWP->query_vars['estate_id'] ?? 0, $pWP->query_vars['view'] ?? '');
 			/* @var $pPdfDownload PdfDownload */
 			$pPdfDownload = $pDI->get(PdfDownload::class);
+            $pPdfDownload->settingGoogleBotAcceptIndex(!empty(get_option('onoffice-settings-google-bot-index-pdf-expose')));
```
jayay, line 172: `PdfDownload` should read that option, so that reading the option can be tested as well.

plugin/PDF/PdfDownload.php
```diff
@@ -65,4 +73,15 @@ public function download(PdfDocumentModel $pModel)
...
 	}
+
+    /**
+     * @param bool $accept
+     */
+    public function settingGoogleBotAcceptIndex(bool $accept = true)
+    {
+        if ($accept === false)
+        {
+            $this->_httpHeadersGeneric->addHeader('X-Robots-Tag: googlebot: noindex, nofollow');
```
jayay, line 84: This: 
```php
$this->_httpHeadersGeneric->addHeader('X-Robots-Tag: googlebot: noindex, nofollow');
```
should be part of the `download` method, That would make the usage more straightforward because that would disallow sequential coupling. In addition to that, it would save us from the boolean parameter.

jayay (changes requested):
One more thing: Please remove the call to `settingGoogleBotAcceptIndex()` from plugin.php, because that's not testable and requires for sequential coupling.

If you add this (untested):
```php
if ($this->_pWPOptionsWrapper->getOption('onoffice-settings-google-bot-index-pdf-expose')) {
    $this->_pHttpHeadersGeneric->addHeader('X-Robots-Tag: googlebot: noindex, nofollow');
}
```
to `PdfDownload::download()` it resolves the issues. `_httpHeadersGeneric` should be renamed into `_pHttpHeadersGeneric`.
Could you please also check your editor settings? The indents look weird in some spots.

plugin/PDF/PdfDownload.php
hungnc89: I updated

plugin.php
hungnc89: I updated

plugin/PDF/PdfDownload.php
```diff
@@ -42,21 +43,28 @@ class PdfDownload
...
-     * @param HTTPHeaders $httpHeadersGeneric
+     * @param HTTPHeaders $pHttpHeadersGeneric
+     * @param WPOptionWrapperDefault $pWPOptionWrapper
      */
 	public function __construct(
 		PdfDocumentFetcher $pPdfDocumentFetcher,
 		PdfDocumentModelValidator $pPdfDocumentModelValidator,
-        HTTPHeaders $httpHeadersGeneric)
+        HTTPHeaders $pHttpHeadersGeneric,
+        WPOptionWrapperDefault $pWPOptionWrapper = null)
```
jayay, line 62: `WPOptionWrapperBase` should be requested here and ` = null` should be removed.

jayay (changes requested):
Looks good so far. I've added another comment. Could you please fix the indents? Those should be tabs, not spaces.

plugin/PDF/PdfDownload.php
hungnc89: I removed it

jayay: `WPOptionWrapperDefault $pWPOptionWrapper = null` should be `WPOptionWrapperBase $pWPOptionWrapper`

hungnc89: I updated it, thanks

jayay (approved):

jayay:
Could you please rebase this on top of the current master?

hungnc89:
> Could you please rebase this on top of the current master?

Ok I will

jayay (approved):

PR #50, opened 2021-02-05
9998 error sent email contact form detail page

hungnc89:


jayay:
Excuse me for the miscommunication. After checking I noticed the correct parameter name is `estateid`. Renaming this probably works because it is unused then.

PR #59, opened 2021-03-04
9877 encrypt api credentials

hungnc89:


tests/TestClassEncrypter.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
```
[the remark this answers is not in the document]
hungnc89: I removed it

plugin/Utility/SymmetricEncryption.php
```diff
@@ -0,0 +1,52 @@
...
+			return $plainText;
+		}
+
+		$ivlen = openssl_cipher_iv_length(self::CIPHER);
+		$iv = openssl_random_pseudo_bytes($ivlen);
+		if ($iv === false) {
+			throw new \RuntimeException(self::IV_ERROR_MESSAGE);
+		}
+		$cipherText = openssl_encrypt($plainText, self::CIPHER, self::$key, OPENSSL_RAW_DATA, $iv);
+		$hmac = hash_hmac('sha256', $cipherText, self::$key, $as_binary = true);
```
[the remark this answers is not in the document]
hungnc89: I added iv to the HMAC

plugin/Utility/SymmetricEncryption.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+
+namespace onOffice\WPlugin\Utility;
+
+
+class SymmetricEncryption
+{
+	const CIPHER = 'AES-128-CBC';
+	private static $key = ONOFFICE_CREDENTIALS_ENC_KEY;
```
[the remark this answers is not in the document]
hungnc89: Update to non-static and type hint

plugin/Utility/SymmetricEncryption.php
```diff
@@ -6,44 +6,55 @@
...
-	public static function encrypt($plainText): string
+	/**
+	 * @param string $plainText
+	 * @param string $key
+	 * @param string $cipher
+	 * @return string
+	 */
+	public function encrypt(string $plainText, string $key, string $cipher = 'AES-128-CBC'): string
 	{
 		if (!extension_loaded('openssl')) {
```
jayay, line 20: Now that the `SymmetricEncryption` class is more generic, just returning the `$plainText` is dangerous. An Exception indicating the OpenSSL extension is not available could be thrown and the fallback case could then be handled externally.

jayay (changes requested):
This is a lot better in terms of reusability. I would appreciate if the `SymmetricEncryption` object could be obtained using dependency injection, instead of with `new`. In addition to that, there is one important thing missing: `ONOFFICE_CREDENTIALS_ENC_KEY` is not necessarily defined. Another class around `SymmetricEncryption` might be needed that checks if

1. OpenSSL extension is installed. If not, skip encryption/decryption (this check currently exists in SymmetricEncryption and should be removed from threre.)
2. check if `ONOFFICE_CREDENTIALS_ENC_KEY` is defined and not empty. If it is undefined or empty, the encryption/decryption should be skipped as well

Otherwise it should call SymmetricEncryption for the encryption/decryption.

plugin/Utility/SymmetricEncryption.php
hungnc89: ok I will fix that. Thank for your response.

hungnc89: I updated as you mentioned: 
 - If the ONOFFICE_CREDENTIALS_ENC_KEY does not exist, It will skipped the encryption.
 - If the OpenSSL is not installed. It will throw an exception.

jayay (approved):
Thank you! I think this is ready to get tested then.

hungnc89:
@jayay please re check this PR. I updated catch exception when cannot decrypt. If we don't catch this will cause error while old secret that haven't ecrypted yet so it cannot be decrypted.

jayay:
Looks good.

jayay (approved):

hungnc89:
@jayay I updated the code so the api key will be encrypted too. Can you take a look?

plugin/Gui/AdminPageApiSettings.php
```diff
@@ -32,6 +32,7 @@
 use function get_option;
 use function json_encode;
 use onOffice\WPlugin\Utility\SymmetricEncryption;
+use PHP_CodeSniffer\Exceptions\RuntimeException;
```
jayay, line 35: Please use PHP's `\RuntimeException` here

jayay: Or remove the `use` import, since it's not in use anyway

hungnc89: I removed it

jayay (approved):

PR #62, opened 2021-03-17
11146 remove options do not show  reference and archived

hungnc89:


jayay (approved):
Thank you, this looks good. I'll try to find out what's wrong with the tests.

PR #67, opened 2021-03-29
11119-visible faktura fields

hungnc89:


plugin/Field/Collection/FieldsCollectionToContentFieldLabelArrayConverter.php
```diff
@@ -48,6 +48,9 @@ public function convert(FieldsCollection $pFieldsCollection, string $module): ar
 			$content = $pField->getCategory() ?: __('(unnamed category)', 'onoffice-for-wp-websites');
 
 			$categories []= $content;
+			if ($key === 'addressFakturaSeparator1' || $key === 'addressFakturaSeparator2') {
```
hungnc89, line 51: I think we only need to check if that fields (Faktura) then we don't converter to "ContentFieldLabelArray"

I do not understand what this requirement means?
"The filter expression" Structure.Table = addressFaktura "can be used to query all billing fields. This filter is to be used avoid displaying these fields."

jayay: Yes. I think the best way to solve this would be in `FieldLoaderGeneric`. Something like this

```php
if ($fieldProperties['module'] === onOfficeSDK::MODULE_ADDRESS && 
    $fieldProperties['tablename'] === 'addressFaktura') {
    continue;
}

jayay (changes requested):
Since this should always apply, I think `FieldLoaderGeneric` would be the best spot. It is important to filter by table name, not field, since there is a lot of faktura fields and individual ones can be created.

jayay (approved):
Thank you! LGTM 👍🏻

PR #68, opened 2021-03-29
#11122 configuration of pagination

hungnc89:


jayay (approved):
Looks good. Can this be tested using unit or integration tests?

hungnc89:
@jayay I updated the paging function to the templates. Can you have a look?

jayay:
Looks good to me. Regarding the test: do you think it is possible to check if the pagination works if a new test was added to `TestTemplateEstateDefaultDetail`?

hungnc89:
> Looks good to me. Regarding the test: do you think it is possible to check if the pagination works if a new test was added to `TestTemplateEstateDefaultDetail`?

@jayay let me check it and reply

hungnc89:
> Looks good to me. Regarding the test: do you think it is possible to check if the pagination works if a new test was added to `TestTemplateEstateDefaultDetail`?

Hi @jayay, I don't add pagination to default detail template because my current theme using it. It does not have pagination. The pagination only appears in other pages.

PR #75, opened 2021-04-28
11300 radius search does not work

hungnc89:


plugin/Controller/EstateListBase.php
```diff
@@ -126,6 +127,23 @@ public function getGeoSearchBuilder(): GeoSearchBuilder;
...
+	public function getGeoSearchBuilderSimilarEstates(): GeoSearchBuilderSimilarEstates;
+
+
+	/**
+	 *
+	 * @param GeoSearchBuilderSimilarEstates $pGeoSearchBuilder
+	 *
+	 */
+
+	public function setGeoSearchBuilderSimilarEstates(GeoSearchBuilderSimilarEstates $pGeoSearchBuilder);
```
jayay, line 145: Why cant `setGeoSearchBuilder` be used?

hungnc89: I can't use "setGeoSearchBuilder".Because I don't know where and how it was used. I'm afraid I'll overwrite your old code.
If you still want to use it I will fix it

PR #79, opened 2021-05-04
11300-radius-not-work-similar-estate

hungnc89:


jayay:
Looks good. Any change this can be tested?

plugin/EstateDetail.php
```diff
@@ -144,16 +151,25 @@ public function getMovieEmbedPlayers(array $options = []): array
...
+	 * @throws API\APIEmptyResultException
+	 * @throws API\ApiClientException
+	 * @throws DataView\UnknownViewException
+	 * @throws DependencyException
+	 * @throws NotFoundException
+	 * @throws HttpFetchNoResultException
 	 */
 
-	public function getSimilarEstates(): string
+	public function getSimilarEstates($pContainer): string
```
jayay, line 164: This would break since it is being called in the `default_detail` template.

jayay (changes requested):
Changing the signature of `\onOffice\WPlugin\EstateDetail::getSimilarEstates` might break the templates

plugin/EstateDetail.php
jayay: It could be an optional parameter

jayay:
Is there any news on this?

nglelinh:
The discussion about this PR is here https://github.com/orgs/onOfficeGmbH/teams/oo-wp/discussions/16
We mentioned that with current structure we could not write unittest for the feature. @jayay  Could you please consider again?

jayay (approved):

PR #81, opened 2021-05-17
12475 add curl option connect timeout 5s

hungnc89:


plugin/SDKWrapperAddCurlOption.php
```diff
@@ -0,0 +1,73 @@
...
+use onOffice\SDK\onOfficeSDK;
+use onOffice\WPlugin\WP\WPOptionWrapperBase;
+use onOffice\WPlugin\WP\WPOptionWrapperDefault;
+
+
+/**
+ *
+ */
+
+class SDKWrapperAddCurlOption extends SDKWrapper
```
jayay, line 40: What exactly is this class used for?

jayay (commented):
I haven't quite understood the approach you're taking on this task.

plugin/SDKWrapperAddCurlOption.php
hungnc89: @jayay this is an extension with specific parameter on CURL options to verify the latency of the API

jayay:
This is a lot easier, but can be simplified even more. For example, instead of `SDKWrapper::sendRequestsWithCustomCurlOption`, `SDKWrapper` could have a member variable that contains the curl options the way they are right now. This is my (even simplier idea) untested, and as a diff:

```diff
diff --git a/plugin/API/APIClientActionGeneric.php b/plugin/API/APIClientActionGeneric.php
index ffae04e..a46a4a0 100644
--- a/plugin/API/APIClientActionGeneric.php
+++ b/plugin/API/APIClientActionGeneric.php
@@ -102,15 +102,6 @@ class APIClientActionGeneric
                $this->_pSDKWrapper->sendRequests();
        }

-       /**
-        * call addRequestToQueue() first
-        * @param $options
-        */
-       public function sendRequestsWithCustomCurlOption($options)
-       {
-               $this->_pSDKWrapper->sendRequestsWithCustomCurlOption($options);
-       }
-
        /**
         * @return bool
         */
diff --git a/plugin/Cache/CacheHandler.php b/plugin/Cache/CacheHandler.php
index 6ab8f9d..dddfb6a 100644
--- a/plugin/Cache/CacheHandler.php
+++ b/plugin/Cache/CacheHandler.php
@@ -38,9 +38,6 @@ class CacheHandler
        /** @var SDKWrapper */
        private $_pSDKWrapper = null;

-       /** @var SDKWrapperAddCurlOption */
-       private $_pSDKWrapperAddCurlOption = null;
-

        /**
         *
@@ -72,13 +69,12 @@ class CacheHandler

        public function clean()
        {
-               $options = [
+               $pApiCall = new APIClientActionGeneric($this->_pSDKWrapper->withCurlOptions([
                        CURLOPT_SSL_VERIFYPEER => true,
                        CURLOPT_PROTOCOLS => CURLPROTO_HTTPS,
                        CURLOPT_CONNECTTIMEOUT => 1,
-               ];
-               $pApiCall = new APIClientActionGeneric($this->_pSDKWrapper, onOfficeSDK::ACTION_ID_READ,'estate');
-               $pApiCall->addRequestToQueue()->sendRequestsWithCustomCurlOption($options);
+               ]), onOfficeSDK::ACTION_ID_READ,'estate');
+               $pApiCall->addRequestToQueue()->sendRequests();
                $records = $pApiCall->getResultRecords();
                if(!empty($records)){
                        foreach ($this->_pSDKWrapper->getCache() as $pCache) {
diff --git a/plugin/SDKWrapper.php b/plugin/SDKWrapper.php
index f9a918a..fac85e8 100644
--- a/plugin/SDKWrapper.php
+++ b/plugin/SDKWrapper.php
@@ -72,35 +72,15 @@ class SDKWrapper
                $this->_caches = [
                        new DBCache(['ttl' => 3600]),
                ];
-               $config = $this->readConfig();
                $this->_pSDK->setCaches($this->_caches);
-               $this->_pSDK->setApiServer($config['server']);
-               $this->_pSDK->setApiVersion($config['apiversion']);
-               $this->_pSDK->setApiCurlOptions($config['curl_options']);
+               $this->_pSDK->setApiServer('https://api.onoffice.de/api/');
+               $this->_pSDK->setApiVersion('latest');
+               $this->_pSDK->setApiCurlOptions([
+                       CURLOPT_SSL_VERIFYPEER => true,
+                       CURLOPT_PROTOCOLS => CURLPROTO_HTTPS,
+               ]);
        }

-
-       /**
-        *
-        * @return array
-        *
-        */
-
-       private function readConfig(): array
-       {
-               $localconfig = [
-                       'apiversion' => 'latest',
-                       'server' => 'https://api.onoffice.de/api/',
-                       'curl_options' => [
-                               CURLOPT_SSL_VERIFYPEER => true,
-                               CURLOPT_PROTOCOLS => CURLPROTO_HTTPS,
-                       ],
-               ];
-
-               return $localconfig;
-       }
-
-
        /**
         *
         * @param APIClientActionGeneric $pApiAction
@@ -147,32 +127,6 @@ class SDKWrapper
                $this->_callbacksAfterSend = [];
        }

-
-       /**
-        * @param $options
-        */
-
-       public function sendRequestsWithCustomCurlOption($options)
-       {
-               $cloneOnOfficeSDK = clone $this->_pSDK;
-               $cloneOnOfficeSDK->setApiCurlOptions($options);
-
-               $pOptionsWrapper = $this->_pWPOptionWrapper;
-               $token = $pOptionsWrapper->getOption('onoffice-settings-apikey');
-               $secret = $pOptionsWrapper->getOption('onoffice-settings-apisecret');
-               $cloneOnOfficeSDK->sendRequests($token, $secret);
-               $errors = $cloneOnOfficeSDK->getErrors();
-
-               foreach ($this->_callbacksAfterSend as $handle => $callback) {
-                       $response = $cloneOnOfficeSDK->getResponseArray($handle) ?? $errors[$handle] ?? [];
-                       call_user_func($callback, $response);
-               }
-
-               $this->_callbacksAfterSend = [];
-       }
-
-
-
        /**
         *
         * @return onOfficeSDKCache[]
@@ -207,4 +161,20 @@ class SDKWrapper
        {
                return $this->_pWPOptionWrapper;
        }
+
+       /**
+        * @param array $curlOptions
+        */
+       public function withCurlOptions(array $curlOptions) : self
+       {
+               $pClone = clone $this;
+               $pClone->_pSDK->setApiCurlOptions($curlOptions);
+               return $pClone;
+       }
+
+       public function __clone()
+       {
+               $this->_pSDK = clone $this->_pSDK;
+               $this->_callbacksAfterSend = [];
+       }
 }
```

tests/SDKWrapperMocker.php
```diff
@@ -182,6 +183,17 @@ public function getRequestArray(): array
 
 
 	/**
+	 *
+	 * @param onOfficeSDK $officeSDK
+	 */
+
+	public function setSDK(onOfficeSDK $officeSDK)
```
jayay, line 190: The new setter gives the mock a different interface than the class it mocks and requires $this->_pSDK to be protected. Is there a different solution?

plugin/SDKWrapper.php
```diff
@@ -205,4 +188,30 @@ public function getWPOptionWrapper(): WPOptionWrapperBase
...
+	}
+
+
+	/**
+	 *
+	 */
+
+	public function __clone()
+	{
+		if (is_object($this->_pSDK)) {
```
jayay, line 212: how can `$this->_pSDK` not be an object? The variable gets initialized in the constructor...

plugin/Cache/CacheHandler.php
```diff
@@ -68,9 +68,31 @@ public function clear()
...
+			}
 		}
 	}
+
+
+	/**
+	 *
+	 */
+
+	public function setUpApiCallWithCurlOptions(): array
```
jayay, line 84: The method is kind of awkward to test. I think that is because this method does an entirely different thing: it tests for the availability of the API. This has nothing to do with the CacheHandler, except that CacheHandler is going to call it. 

=> Please create a new class (for example `APIAvailabilityChecker`) and move the method there. The method should get a new name as well, for example `checkAvailability()`.

plugin/SDKWrapper.php
```diff
@@ -45,7 +45,7 @@
 class SDKWrapper
 {
 	/** @var onOfficeSDK */
-	private $_pSDK = null;
+	protected $_pSDK = null;
```
jayay, line 48: Please make this private again

jayay (changes requested):
Please see my comments. The method that checks for the availability of the API should be in its own class. That way, the test for CacheHandler can be cleaned up and SDKWrapperMocker as well. If you need to be able to set the API URL, please create a new "with" method in SDKWrapper.

plugin/Cache/CacheHandler.php
hungnc89: In which folder should I create new class?

plugin/Cache/CacheHandler.php
```diff
@@ -68,9 +68,31 @@ public function clear()
 
 	public function clean()
 	{
-		foreach ($this->_pSDKWrapper->getCache() as $pCache) {
-			/* @var $pCache onOfficeSDKCache */
-			$pCache->cleanup();
+		if(!empty($this->setUpApiCallWithCurlOptions())){
```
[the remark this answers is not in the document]
hungnc89: That's right, so I'm trying to read basicsettings instead of estates.

plugin/API/APIAvailabilityChecker.php
```diff
@@ -0,0 +1,47 @@
...
+use onOffice\WPlugin\SDKWrapper;
+
+class APIAvailabilityChecker
+{
+	/**
+	 *
+	 * @throws ApiClientException
+	 */
+
+	public function checkAvailability(SDKWrapper $pSDKWrapper): array
```
jayay, line 18: `SDKWrapper` can be injected in constructor.

plugin/API/APIAvailabilityChecker.php
```diff
@@ -0,0 +1,47 @@
...
+							"color2",
+							"textcolorMail",
+							"claim"
+						],
+					]
+				]
+			]
+		);
+		$pApiCall->addRequestToQueue()->sendRequests();
+		return $pApiCall->getResultRecords();
```
jayay, line 45: Why not return a boolean?

jayay: If this method will return a boolean, it would be better if the method name was `isAvailable` or something similar

jayay (changes requested):
This looks quite good. I've noticed two minor things in terms of architecture.

plugin/API/APIAvailabilityChecker.php
```diff
@@ -0,0 +1,61 @@
...
+{
+	/** @var SDKWrapper */
+	private $_pSDKWrapper = null;
+
+	/**
+	 *
+	 * @param SDKWrapper $pSDKWrapper
+	 */
+
+	public function __construct(SDKWrapper $pSDKWrapper = null)
```
jayay, line 21: ` = null` and `?? new SDKWrapper` aren't necessary, the dependency gets resolved using autowiring

jayay (changes requested):

jayay (approved):
Thank you!

PR #82, opened 2021-05-19
12764 remove backend data types from configuration mask that do not make sense

hungnc89:


hungnc89:
closed due to API does not support right now

PR #85, opened 2021-06-10
#13585 remove contact person default unit template

hungnc89:


hungnc89:
@jayay Can you help me why the unit test failed when I created the PR but it does not fail when I push the code. I cannot find the unit test that failed in the code

jayay (approved):

PR #87, opened 2021-06-16
13584 update status estate detail

hungnc89:


jayay:
Does this work at all? Because the field `vermarktungsstatus` does not get delivered by the API. Instead, it needs to be generated by the plugin in the same way as it is being done for the list view: https://github.com/onOfficeGmbH/oo-wp-plugin/blob/e9ab3cd1b0484577fe561e65efb0d61bbd87a6ef/plugin/EstateList.php#L389

hungnc89:
> Does this work at all? Because the field `vermarktungsstatus` does not get delivered by the API. Instead, it needs to be generated by the plugin in the same way as it is being done for the list view:
> 
> https://github.com/onOfficeGmbH/oo-wp-plugin/blob/e9ab3cd1b0484577fe561e65efb0d61bbd87a6ef/plugin/EstateList.php#L389

I tested it worked locally 
image

jayay (approved):
Thank you!

PR #95, opened 2021-07-22
update fix error region id is null

hungnc89:


plugin/Region/RegionController.php
```diff
@@ -100,6 +100,9 @@ private function createRegionObjectsRegionOne(array $regionList): array
 	{
 		$regions = [];
 		foreach ($regionList as $regionProperties) {
+			if (empty($regionProperties['elements']['id'])) {
```
jayay, line 103: `!isset()` would be preferred over `empty()`. That's a minor thing, though.

jayay (commented):
Can this change be tested? That'd be good

plugin/Region/RegionController.php
hungnc89: `empty()` includes checking that the variable exists and is not null. If the api returns id is null there will still be an error

hungnc89:
> Can this change be tested? That'd be good

hungnc89:
Hi @jayay,
I added test for the task

jayay:
Thank you!

PR #97, opened 2021-08-02
P#50199 show fields message and comment on interest forms

hungnc89:


plugin/Field/SearchcriteriaFields.php
```diff
@@ -108,8 +108,11 @@ public function getFieldLabelsOfInputs(array $inputFormFields): array
 		$output = [];
 		$pGeoPosition = new GeoPosition;
 		$geoRangeFields = array_flip($pGeoPosition->getSearchCriteriaFields());
-
 		foreach ($inputFormFields as $name => $value) {
+			if (in_array($name, ['message'])) {
```
jayay, line 112: I think, instead of adding new ifs and bool parameters, it should be found out why `$pFieldsCollection->getFieldByModuleAndName()` doesn't know the `message` field anymore

plugin/FormData.php
```diff
@@ -162,7 +163,7 @@ public function getSearchcriteriaData(): array
 			$inputConfigName = $pSearchcriteriaFields->getFieldNameOfInput($input);
 			$inputModule = $inputs[$inputConfigName] ?? null;
 
-			if (onOfficeSDK::MODULE_SEARCHCRITERIA === $inputModule) {
+			if (onOfficeSDK::MODULE_SEARCHCRITERIA === $inputModule || ($isGetMessageField && in_array($input, ['message']))) {
```
jayay, line 166: See comment above

jayay (changes requested):
I think what's missing is simply that [FieldModuleCollectionDecoratorSearchcriteria](https://github.com/onOfficeGmbH/oo-wp-plugin/blob/0ff2b07561bad6b46d9e4edaa59926807f34b560/plugin/Field/FieldModuleCollectionDecoratorSearchcriteria.php) needs to know about the `message` field.

plugin/Field/SearchcriteriaFields.php
hungnc89: Hi @jayay. I do not understand. Because the `message` field is not returned by the API. Which module do I need to add the `message` field to?

hungnc89:
@jayay please see comment above

plugin/Field/SearchcriteriaFields.php
jayay: Those were usually added to the `''` (empty string) module.

hungnc89: I add a message field to the search criteria module. But when submitting the form, there will be an error. Because it will call CreateSearchCriteria API and will give error no message

jayay: This change worked for me:

```diff
diff --git a/plugin/Field/SearchcriteriaFields.php b/plugin/Field/SearchcriteriaFields.php
index 57754ef..87c362b 100644
--- a/plugin/Field/SearchcriteriaFields.php
+++ b/plugin/Field/SearchcriteriaFields.php
@@ -102,29 +102,24 @@ class SearchcriteriaFields
                        ->addFieldsSearchCriteria($pFieldsCollection)
                        ->addFieldsAddressEstate($pFieldsCollection);

-               $pGeoFieldsCollection = new FieldModuleCollectionDecoratorGeoPositionFrontend(new FieldsCollection);
-               $pFieldsCollection->merge($pGeoFieldsCollection);
+               $pFieldsCollection->merge(new FieldModuleCollectionDecoratorGeoPositionFrontend(new FieldsCollection));
+               $pFieldsCollection->merge(new FieldModuleCollectionDecoratorFormContact(new FieldsCollection));

                $output = [];
                $pGeoPosition = new GeoPosition;
                $geoRangeFields = array_flip($pGeoPosition->getSearchCriteriaFields());
                foreach ($inputFormFields as $name => $value) {
-                       if (in_array($name, ['message'])) {
-                               $output[$name] = $value;
-                               continue;
-                       }
                        $aliasedFieldName = $this->getFieldNameOfInput($name);
                        if (in_array($name, $pGeoPosition->getSearchCriteriaFields())) {
                                $aliasedFieldName = $geoRangeFields[$name];
                        }
-                       $module = onOfficeSDK::MODULE_ESTATE;
-                       $pField = $pFieldsCollection->getFieldByModuleAndName($module, $aliasedFieldName);
+                       $pField = $pFieldsCollection->getFieldByKeyUnsafe($aliasedFieldName);

                        if (FieldTypes::isRangeType($pField->getType()))
                        {
-                               if (stristr($name, self::RANGE_FROM)) {
+                               if (__String::getNew($name)->endsWith(self::RANGE_FROM)) {
                                        $output[$pField->getLabel().' (min)'] = $value;
-                               } elseif (stristr($name, self::RANGE_UPTO)) {
+                               } elseif (__String::getNew($name)->endsWith(self::RANGE_UPTO)) {
                                        $output[$pField->getLabel().' (max)'] = $value;
                                } else {
                                        $output[$pField->getLabel()] = $value;
diff --git a/templates.dist/form/applicantform.php b/templates.dist/form/applicantform.php
index 5b6076d..a1fa5e0 100644
--- a/templates.dist/form/applicantform.php
+++ b/templates.dist/form/applicantform.php
@@ -58,6 +58,10 @@ foreach ( $pForm->getInputFields() as $input => $table ) {
                $searchcriteriaValues []= $line;
        }

+       if ($table === ''){
+               $addressValues []= $line;
+    }
+
 }

 if ($pForm->getFormStatus() !== \onOffice\WPlugin\FormPost::MESSAGE_SUCCESS) {
```

Calling `$pFieldsCollection->getFieldByKeyUnsafe()` is good enough for now, but could lead to strange issues later. It is better to know the module and call `getFieldByModuleAndName`, but the module of our message field is `''` and it'll make it harder later if the code has a bunch of `if ($field == 'message')` special cases.

plugin/FormData.php
jayay, line 166: I would say the `message` field is neither a search criteria field, nor address or estate. It is something special that is only supposed to be used in forms.

jayay (commented):
I answered some questions. See above

hungnc89:
@jayay I updated it

jayay:
Thanks! Could you please look after the tests?

hungnc89:
@jayay I update it

jayay:
Didn't 9169e98a2c9ec743e06429cc272457cb58854d77 remove the test for the form field altogether?

hungnc89:
@jayay That test I wrote did not match your requirements so it should be removed

michaelarnold-dev:
@tung-le-esg Could you please check this as well?

tung-le-esg:
> @tung-le-esg Could you please check this as well?

Yes, of course

jayay (approved):

michaelarnold-dev:
@nglelinh : according to Robert Igelmund that has not been fixed so far. He prepared a video of the bug

https://user-images.githubusercontent.com/77264788/149947607-a512da0a-282c-437f-95b2-c437e6c6e7a0.mp4

hoangdoegs:
@michaelarnold-dev I fixed the error. Please check again. Thanks

michaelarnold-dev:
@jayay please see the comment of hoangdoegs

hungnc89:
@jayay The tests have been added from #171

jayay (approved):

michaelarnold-dev:
@nglelinh according to Robert Igelmund one point is still wrong:  When creating a new interest search form, the "Nachricht" field is selectable in the backend.
tested in Version 2.20.12-13-g9b7f73c2

nglelinh:
@michaelarnold-dev seems that in the requirement we should care about fields “message” and “comment” on interest forms only

jayay:
@nglelinh Yes, please also fix the application search form.

plugin/Field/Collection/FieldsCollectionToContentFieldLabelArrayConverter.php
```diff
@@ -45,6 +47,9 @@ public function convert(FieldsCollection $pFieldsCollection, string $module): ar
 		$categories = [];
 
 		foreach ($pFieldsCollection->getFieldsByModule($module) as $key => $pField) {
+			if (!$this->_showMessageField && $key === 'message') {
```
jayay, line 50: `$_showMessageField` does not belong in this class. Field-specific logic belongs into the classes that build the `FieldsCollection`. This class is supposed to remain stateless.

jayay (changes requested):

plugin/Field/Collection/FieldsCollectionToContentFieldLabelArrayConverter.php
hoangdoegs: I am having trouble using `removeFieldByModuleAndName` when the module for the Message field is null. I need to check when in Applicant Search Form will remove Message field

jayay: Doesn't an empty string work for the module then?

hoangdoegs: Yes, the empty module doesn't work with `removeFieldByModuleAndName`

michaelarnold-dev: @jayay Can you help here?

jayay: `$pFieldsCollection->removeFieldByModuleAndName('', $fieldName);` should be a perfectly valid invocation (if the module is an empty string, not `null`).

jayay: But I think `message` shouldn't end up in the `FieldsCollection` in the first place.

plugin/Field/Collection/FieldsCollectionToContentFieldLabelArrayConverter.php
```diff
@@ -31,6 +31,8 @@
 
 class FieldsCollectionToContentFieldLabelArrayConverter
 {
+
+	private $_showMessageField = true;
```
jayay, line 35: This class needs to stay stateless, and I don't think this is in use at all, so please remove this

plugin/Field/SearchcriteriaFields.php
```diff
@@ -102,13 +102,13 @@ public function getFieldLabelsOfInputs(array $inputFormFields): array
 			->addFieldsSearchCriteria($pFieldsCollection)
 			->addFieldsAddressEstate($pFieldsCollection);
 
-		$pGeoFieldsCollection = new FieldModuleCollectionDecoratorGeoPositionFrontend(new FieldsCollection);
-		$pFieldsCollection->merge($pGeoFieldsCollection);
+		$pFieldsCollection->merge(new FieldModuleCollectionDecoratorGeoPositionFrontend(new FieldsCollection));
```
jayay, line 105: This might have a major impact. I hope this works out to be well. The FieldModuleCollectionDecorator* classes were supposed to be decorating the `FieldsCollection` with additional (fictional) fields for a particular purpose, for example the use in a contact form (message is one of those fictional fields). This might unleash fields in areas where they're not supposed to be.

jayay: This is the reason why `if ($this->_showMessageField)` is needed down below.

plugin/Gui/AdminPageFormSettingsBase.php
```diff
@@ -108,6 +108,9 @@ abstract class AdminPageFormSettingsBase
 	/** @var FormModelBuilderDBForm */
 	private $_pFormModelBuilder = null;
 
+	/** @var bool */
+	private $_showMessageField = true;
```
jayay, line 112: See above

plugin/Field/SearchcriteriaFields.php
jayay: I think a new `FieldModuleCollectionDecoratorInterestForms` would be the better solution

plugin/Installer/DatabaseChanges.php
```diff
@@ -186,6 +186,10 @@ public function install()
 			dbDelta($this->getCreateQueryForms());
 			$dbversion = 25;
 		}
+		if ($dbversion == 25) {
+			dbDelta($this->deleteMessageFieldApplicantSearchForm());
```
jayay: no `dbDelta` needed here, just the method call

plugin/Installer/DatabaseChanges.php
```diff
@@ -223,6 +227,24 @@ private function getCreateQueryCache()
...
+		$prefix = $this->getPrefix();
+		$tableName = $prefix . "oo_plugin_forms";
+		$tableFieldConfig = $prefix . "oo_plugin_form_fieldconfig";
+
+		$rows = $this->_pWPDB->get_results("SELECT `form_id` FROM {$tableName} WHERE form_type = 'applicantsearch'");
+
+		foreach ($rows as $applicantSearchFormId) {
+			$allFieldMessages = $this->_pWPDB->get_results("SELECT form_fieldconfig_id FROM $tableFieldConfig
+										WHERE `fieldname` = 'message'
+										AND `form_id` = '{$applicantSearchFormId->form_id}'");
```
jayay, line 241: Please add proper escaping here (`$applicantSearchFormId->form_id`) and please also rename `$applicantSearchFormId`

plugin/Field/SearchcriteriaFields.php
```diff
@@ -119,9 +119,9 @@ public function getFieldLabelsOfInputs(array $inputFormFields): array
 
 			if (FieldTypes::isRangeType($pField->getType()))
 			{
-				if (__String::getNew($name)->endsWith(self::RANGE_FROM)) {
```
jayay: Does this do the same? If so, I would prefer the `__String::getNew($name)->endsWith(self::RANGE_UPTO)` variant.

jayay (changes requested):

plugin/Installer/DatabaseChanges.php
```diff
@@ -630,10 +630,10 @@ private function deleteCommentFieldApplicantSearchForm()
 
 		$rows = $this->_pWPDB->get_results("SELECT `form_id` FROM {$tableName} WHERE form_type = 'applicantsearch'");
 
-		foreach ($rows as $applicantSearchFormId) {
+		foreach ($rows as $applicantSearchForm) {
 			$allFieldComments = $this->_pWPDB->get_results("SELECT form_fieldconfig_id FROM $tableFieldConfig
 										WHERE `fieldname` = 'krit_bemerkung_oeffentlich'
-										AND `form_id` = '{$this->_pWPDB->_escape($applicantSearchFormId->form_id)}'");
+										AND `form_id` = '{$this->_pWPDB->_escape($applicantSearchForm->form_id)}'");
```
jayay, line 636: same here

jayay (changes requested):

plugin/FormPostInterest.php
```diff
@@ -138,7 +137,6 @@ private function sendEmail(FormData $pFormData, string $recipient, $subject = nu
 			'formtype' => $pFormData->getFormtype(),
 			'referrer' => filter_input(INPUT_SERVER, 'REQUEST_URI') ?? '',
 			'recipient' => $recipient,
-			'krit_bemerkung_oeffentlich' => $comment,
```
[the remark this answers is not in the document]
vu-vuong-esg: because, field 'krit_bemerkung_oeffentlich' inside variable $searchCriterias, therefore i removed at here

vu-vuong-esg:
@jayay , i fixed PR, can you check help me ?

jayay (approved):

jmaas-onoffice (approved):

PR #99, opened 2021-08-06
P#50175 remove approx when displaying area values

vu-vuong-esg:


jayay:
Thank you! Did you see that `EstateList` has the values both pre-formatted (that's what the templates get in the loop) and as raw values? The raw values aren't usable outside the class. Together with a check which table the field is from, general rules about formatting could be established. The newly formatted value could be written back into `EstateList::$_records`.

See https://github.com/onOfficeGmbH/oo-wp-plugin/blob/b7d24f6754020f9785677aaab2ec9f794ed9b1e4/plugin/EstateList.php#L197-L213

vu-vuong-esg:
> Thank you! Did you see that `EstateList` has the values both pre-formatted (that's what the templates get in the loop) and as raw values? The raw values aren't usable outside the class. Together with a check which table the field is from, general rules about formatting could be established. The newly formatted value could be written back into `EstateList::$_records`.
> 
> See
> 
> https://github.com/onOfficeGmbH/oo-wp-plugin/blob/b7d24f6754020f9785677aaab2ec9f794ed9b1e4/plugin/EstateList.php#L197-L213

Actually I don't touch the record raws. I updated the $recordModified variable instead 
>https://github.com/onOfficeGmbH/oo-wp-plugin/blob/b7d24f6754020f9785677aaab2ec9f794ed9b1e4/plugin/EstateList.php#L385
If we touch $this->_records we need to loop through it before any iterator. So I think I will format it when iterator is better.

vu-vuong-esg:
Hi @jayay, can you give a check. l used regex for checking value and reformat value

jayay:
The `&` operators in the loops make the code hard to maintain. It is really not necessary to use string functions to modify the values. Why not do it like this:

```php
foreach ($this->_records as $record) {
    foreach ($record['elements'] ?? [] as $fieldName => $value) {
        if (FieldTypes::isNumeric($this->_pEnvironment->getFieldnames()->getType($fieldName, \onOffice\SDK\onOfficeSDK::MODULE_ESTATE))) {
            $this->_records[$record][$fieldName] = $this->_recordsRaw[$record][$fieldName];
        }
    }
}
```

vu-vuong-esg:
> The `&` operators in the loops make the code hard to maintain. It is really not necessary to use string functions to modify the values. Why not do it like this:
> 
> ```
> foreach ($this->_records as $record) {
>     foreach ($record['elements'] ?? [] as $fieldName => $value) {
>         if (FieldTypes::isNumeric($this->_pEnvironment->getFieldnames()->getType($fieldName, \onOffice\SDK\onOfficeSDK::MODULE_ESTATE))) {
>             $this->_records[$record][$fieldName] = $this->_recordsRaw[$record][$fieldName];
>         }
>     }
> }
> ```

I updated it, can you take a look @jayay

PR #105, opened 2021-08-13
P#51787 update fix display parking

hungnc89:
Related to issue #295 https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/374

changed logs:
-


plugin/EstateList.php
```diff
@@ -656,6 +661,86 @@ public function getEstateIds(): array
...
+	public function formatParkingLot(array $parkingArray, string $language, string $locate = 'de_DE'): array
+	{
+		$messages = [];
+		foreach ($parkingArray as $key => $parking) {
+			if (!$parking['Count']) {
+				continue;
+			}
+			switch ($language) {
+				case 'ENG':
+					$element = "{$parking['Count']} {$this->getParkingName($key, $parking['Count'])} at {$this->formatPrice($parking['Price'], $language, $locate)}";
```
jayay, line 679: The string should be translatable in `sprintf()` format.

plugin/EstateList.php
```diff
@@ -656,6 +661,86 @@ public function getEstateIds(): array
 		return array_column($this->_records, 'id');
 	}
 
+	/**
+	 * @param array $parkingArray
+	 * @param string $language
+	 * @param string $locate
+	 * @return array
+	 */
+	public function formatParkingLot(array $parkingArray, string $language, string $locate = 'de_DE'): array
```
jayay, line 670: better call it "locale"

plugin/EstateList.php
```diff
@@ -388,7 +388,12 @@ public function estateIterator($modifier = EstateViewFieldModifierTypes::MODIFIE
 			$pEstateStatusLabel = $this->_pEnvironment->getEstateStatusLabel();
 			$recordModified['vermarktungsstatus'] = $pEstateStatusLabel->getLabel($recordRaw);
 		}
-
+		if (isset($recordModified['multiParkingLot'])) {
+			$language = new Language();
+			$languageDefault = Language::getDefault();
+			$locate = $language->getLocale();
+			$recordModified['multiParkingLot'] = $this->formatParkingLot($recordModified['multiParkingLot'], $languageDefault, $locate);
```
jayay, line 395: `formatParkingLot` is some functionality that needs to be available in the templates. `EstateList` should provide the template the raw parking lot data, not the formatted data.

plugin/EstateList.php
```diff
@@ -656,6 +661,86 @@ public function getEstateIds(): array
...
+			return $str;
+		}
+	}
+
+	/**
+	 * @param string $parkingName
+	 * @param int $count
+	 * @return string
+	 */
+	public function getParkingName(string $parkingName, int $count): string
```
jayay, line 735: This highly depends on the language, WP has something for this, but the __* functions cannot be used with variables, but should always have a string literal as the text to be translated. See https://developer.wordpress.org/reference/functions/_n/ and https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#plurals

plugin/EstateList.php
```diff
@@ -656,6 +661,86 @@ public function getEstateIds(): array
...
+	{
+		$digit = intval(substr(strrchr($str, "."), 1));
+		try {
+			$format = new NumberFormatter($locate, NumberFormatter::CURRENCY);
+			if ($digit) {
+				$format->setAttribute(NumberFormatter::MAX_FRACTION_DIGITS, 2);
+			} else {
+				$format->setAttribute(NumberFormatter::MAX_FRACTION_DIGITS, 0);
+			}
+			return str_replace("\xc2\xa0", " ", $format->formatCurrency($str, "EUR"));;
```
jayay, line 710: is the `str_replace` needed here?

jayay (changes requested):
A couple of open questions and locale-specific issues, and the templates should be in charge of the formatting.

plugin/EstateList.php
hungnc89: Hi @jayay . Because `$recordModified` is handled by `$pArrayContainer = new ArrayContainerEscape($recordModified)` the values in `multiParkingLot` if is Array will be converted to "Array" string

plugin/EstateList.php
hungnc89: Because `NumberFormatter` `formatCurrency` will return the character [NBSP], it will not pass when running the unit test. I need replace [NBSP] with space

plugin/EstateList.php
```diff
@@ -735,8 +734,10 @@ public function formatPrice(string $str, string $language, string $locate): stri
 	public function getParkingName(string $parkingName, int $count): string
 	{
 		$str = preg_replace('/([a-z])([A-Z])/', "\\1 \\2", $parkingName);
-		if ($count > 1) {
-			$str .= 's';
+		if (1 === $count) {
```
jayay, line 737: a switch/case statement would be better here, so that it can match on each possible `$parkingName`, so that the strings that need to be translated can be string _literals_ in the code. That is important, otherwise the script that generates the _.pot_ file can't find the strings.

hungnc89: @jayay . I updated it

plugin/EstateList.php
jayay: That's bad. `ArrayContainerEscape` will only handle one level of arrays. Would it be a good solution to make `ArrayContainer(Escape)` to handle arbitrary levels of nested arrays?

plugin/EstateList.php
```diff
@@ -656,6 +661,89 @@ public function getEstateIds(): array
...
+	public function formatParkingLot(array $parkingArray, string $language, string $locale = 'de_DE'): array
+	{
+		$messages = [];
+		foreach ($parkingArray as $key => $parking) {
+			if (!$parking['Count']) {
+				continue;
+			}
+			switch ($language) {
+				case 'ENG':
+					$element = sprintf("%d %s at %s", number_format_i18n($parking['Count']), $this->getParkingName($key, $parking['Count']), $this->formatPrice($parking['Price'], $language, $locale));
```
jayay, line 679: This part belongs into the template. Also, user-defined translations are possible in templates, so that no switch/case for the language is needed.

plugin/EstateList.php
```diff
@@ -656,6 +661,89 @@ public function getEstateIds(): array
...
+
+	/**
+	 * @param string $parkingName
+	 * @param int $count
+	 * @return string
+	 */
+	public function getParkingName(string $parkingName, int $count): string
+	{
+		$str = preg_replace('/([a-z])([A-Z])/', "\\1 \\2", $parkingName);
+		switch ($count) {
```
jayay, line 737: The `_n()` function uses the singular if `$count` is `1` and the plural otherwise. There are languages with different rules and `_n()` respects those.

jayay: The switch/case should look like this in order to be properly translated:
```php
switch ($parkingName) {
    case 'carport':
        /* translators: %s is the amount of carports */
        $str = _n('%s carport', '%s carports', $count, 'onoffice-for-wp-websites');
        break;
    case 'duplex':
         /* translators: %s is the amount of duplexes */
        $str = _n('%s duplex', '%s duplexes', $count, 'onoffice-for-wp-websites');
        break;
     ....
}
return esc_html(sprintf($str, $count));
```

jayay (changes requested):
I wrote a comment on how to write plural translations. It involves more code but follows the KISS principle.

hungnc89:
@jayay I updated it.

plugin/EstateList.php
```diff
@@ -674,14 +674,7 @@ public function formatParkingLot(array $parkingArray, string $language, string $
...
 			}
-			switch ($language) {
-				case 'ENG':
-					$element = sprintf("%d %s at %s", number_format_i18n($parking['Count']), $this->getParkingName($key, $parking['Count']), $this->formatPrice($parking['Price'], $language, $locale));
-					break;
-				default:
-					$element = sprintf("%d %s à %s", number_format_i18n($parking['Count']), $this->getParkingName($key, $parking['Count']), $this->formatPrice($parking['Price'], $language, $locale));
-					break;
-			}
+			$element = sprintf("%s at %s", $this->getParkingName($key, $parking['Count']), $this->formatPrice($parking['Price'], $language, $locale));
```
jayay, line 677: `%1$d` with ongoing numbers should be used here instead of `%s` so that both kinds of phrases become possible:
* _$500 per garage, 5 available_ (I just noticed this is not possible right now, because the number gets inserted at a specific position in the switch/case below)
* _5 garages, $500 each_

The format needs to be wrapped in the `__()` function as well.

hungnc89:
@jayay I updated it

michaelarnold-dev:
@tung-le-esg : Could you check this?

jayay:
> `formatParkingLot` is some functionality that needs to be available in the templates. `EstateList` should provide the template the raw parking lot data, not the formatted data.

This is still open.

plugin/EstateList.php
```diff
@@ -24,6 +24,7 @@
 use DI\ContainerBuilder;
 use DI\DependencyException;
 use DI\NotFoundException;
+use Nette\Neon\Exception;
```
jayay, line 27: Oops, did that get in on accident?

plugin/EstateList.php
```diff
@@ -51,6 +52,7 @@
 use onOffice\WPlugin\ViewFieldModifier\EstateViewFieldModifierTypes;
 use onOffice\WPlugin\ViewFieldModifier\ViewFieldModifierHandler;
 use onOffice\WPlugin\WP\WPQueryWrapper;
+use function DI\string;
```
jayay, line 55: This looks suspicious too

jayay (changes requested):

tung-le-esg:
@hoangdoegs Please check this PR also.

tests/TestClassEstateList.php
```diff
@@ -752,6 +751,9 @@ public function prepareEstateList()
 		]);
 		$this->_pEnvironment->method('getEstateStatusLabel')->willReturn
 			($pEstateStatusLabel);
+		$pEstateListMock = $this->getMockBuilder(EstateList::class)
```
jayay, line 754: This looks unused

templates.dist/fields.php
```diff
@@ -224,4 +224,93 @@ function renderRegionalAddition(string $inputName, array $selectedValue, bool $m
...
+			if (!empty($parking['MarketingType'])) {
+				$element .= ' (' . $parking['MarketingType'] . ')';
+			}
+			array_push($messages, $element);
+		}
+		return $messages;
+	}
+}
+
+if (!function_exists('formatPrice')) {
```
jayay, line 247: this is global namespace, maybe use a less generic name or put the functions into a namespace? Just a thought...

templates.dist/fields.php
```diff
@@ -224,4 +224,93 @@ function renderRegionalAddition(string $inputName, array $selectedValue, bool $m
...
+
+if (!function_exists('renderParkingLot')) {
+	function renderParkingLot(array $parkingArray, string $language, string $locale = 'de_DE'): array
+	{
+		$messages = [];
+		foreach ($parkingArray as $key => $parking) {
+			if (!$parking['Count']) {
+				continue;
+			}
+			$element = sprintf(__('%1$s at %2$s'), getParkingName($key, $parking['Count']), formatPrice($parking['Price'], $language, $locale));
```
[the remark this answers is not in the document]
jayay: This is still open (2nd argument for `__()` is missing)

jayay (changes requested):
Almost done 👍🏻

jayay (commented):
Thank you, this looks good. Would you please consider rebuilding the `onoffice.pot`, `onoffice-*.mo` and `onoffice-*.po` files?

You could use `make pot` and discard the changes to the `languages/onoffice-for-wp-websites*` files.

nglelinh:
@jayay please help to review again, thanks

jayay:
5424e86 looks good, 60f14f0 went wrong (see commit message). Could you please revert 60f14f0?

hungnc89:
> [5424e86](https://github.com/onOfficeGmbH/oo-wp-plugin/commit/5424e86f55db25ddad755e667316db36cea08638) looks good, [60f14f0](https://github.com/onOfficeGmbH/oo-wp-plugin/commit/60f14f01a1311b592da70394af7cfcc41606d177) went wrong (see commit message). Could you please revert [60f14f0](https://github.com/onOfficeGmbH/oo-wp-plugin/commit/60f14f01a1311b592da70394af7cfcc41606d177)?

Ok, I will update, please review again.

hoangdoegs:
@jayay. I fixed it. Please review again

jayay:
e143ee0 looks broken again.

jayay:
@hoangdoegs @hungnc89 could you please check if the new branch https://github.com/onOfficeGmbH/oo-wp-plugin/tree/15324-fix-display-parking-2 contains all the changes? If so, would you please change its name, so that it get associated with this PR?

nglelinh:
> [e143ee0](https://github.com/onOfficeGmbH/oo-wp-plugin/commit/e143ee0dd0970f6a1dfda8b3c8a436dad42ea0ac) looks broken again.

@jayay could you please help to clarify what was broken here ?

jayay:
The commit message shows 

```
� Conflicts:
�	languages/onoffice-for-wp-websites-de_DE.mo
�	languages/onoffice-for-wp-websites-es_ES.mo
�	languages/onoffice-for-wp-websites-it_IT.mo
�	languages/onoffice-for-wp-websites-nl_NL.mo
```

This shouldn't happen.

templates.dist/fields.php
```diff
@@ -224,4 +224,93 @@ function renderRegionalAddition(string $inputName, array $selectedValue, bool $m
...
+
+if (!function_exists('renderParkingLot')) {
+	function renderParkingLot(array $parkingArray, string $language, string $locale = 'de_DE'): array
+	{
+		$messages = [];
+		foreach ($parkingArray as $key => $parking) {
+			if (!$parking['Count']) {
+				continue;
+			}
+			$element = sprintf(__('%1$s at %2$s', 'onoffice'), getParkingName($key, $parking['Count']), formatPriceParking($parking['Price'], $language, $locale));
```
jayay, line 237: `/* translators: ... */` comment is missing here

hoangdoegs:

templates.dist/fields.php
jayay: The comment needs to explain `%1$s` and `%2$s`, which would be the _type of parking lot_ and its _price_, respectively.

jayay (approved):

michaelarnold-dev:
@hoangdoegs Something seems to go wrong in the code: The tester commented the following:

Test with version Version 2.22.0-22-g3e4c5a78 on https://wpptest.onofficeweb.com/.

1.) In my case "Parking" is shown as ARRAY in the property list. See screenshot. 
2.) Why is there a template selectable called "ParkingLot"? How does it work? That is not a requirement. See screenshot.

tang-hien-egs:
> @hoangdoegs Something seems to go wrong in the code: The tester commented the following:
> 
> Test with version Version 2.22.0-22-g3e4c5a78 on https://wpptest.onofficeweb.com/.
> 
> 1.) In my case "Parking" is shown as ARRAY in the property list. See screenshot. 2.) Why is there a template selectable called "ParkingLot"? How does it work? That is not a requirement. See screenshot   .

@michaelarnold-dev , have you update the template in version laster ?, i tested local not issue above

michaelarnold-dev:
@tang-hien-egs Could you tell which template and what version you are talking about? The tester deleted the plugin and installed the project code  onoffice-v2.22.0-22-g3e4c5a78.zip Afterwards the code did not work

michaelarnold-dev:
@hoangdoegs Any update here? Could you please check my comment?

tang-hien-egs:
> @hoangdoegs Any update here? Could you please check my comment?

sorry, I have reproduced the error, I will update as soon as possible

tang-hien-egs:
> @hoangdoegs Any update here? Could you please check my comment?

@michaelarnold-dev , i fixed PR, can you check PR help me ?

css/onoffice-default.css
```diff
@@ -66,4 +66,7 @@ form[data-applicant-form-id] #spinner {
     display: inline;
     width: 40px;
     vertical-align: middle;
+}
+.oo-listinfo .oo-listinfotable .oo-detailslisttd {
```
jayay, line 70: Could you please move this CSS into the template, because this needs to stay backwards compatible and can barely be changed in the future.

jayay (changes requested):
Almost done, just one small thing! 🙂

css/onoffice-default.css
vu-vuong-esg: > Could you please move this CSS into the template, because this needs to stay backwards compatible and can barely be changed in the future.

@jayay , i fixed it, can you check it help me ?

jmaas-onoffice (approved):

templates.dist/estate/parkingLot/ParkingLot.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
jmaas-onoffice, line 1: If you add the parking lot field to an estate list, the CSS will be for the detail view.

GUI

Therefore please remove this file and inline its code into the templates where it is used. Also make sure that the classes are the same as the rest of the fields in that template.

templates.dist/fields.php
```diff
@@ -223,3 +223,94 @@ function renderRegionalAddition(string $inputName, array $selectedValue, bool $m
...
+if (!function_exists('renderParkingLot')) {
+	function renderParkingLot(array $parkingArray, string $language, string $locale = 'de_DE'): array
+	{
+		$messages = [];
+		foreach ($parkingArray as $key => $parking) {
+			if (!$parking['Count']) {
+				continue;
+			}
+			/* translators: 1: Name of parking lot, 2: Price */
+			$element = sprintf(__('%1$s at %2$s', 'onoffice'), getParkingName($key, $parking['Count']), formatPriceParking($parking['Price'], $language, $locale));
```
jmaas-onoffice, line 237: I have updated the requirements:
Please display to different texts depending on whether it is a single item or multiple (the "each" only appears when there is more than one):
- 1 garage at €500 (purchase)
- 10 free parking lots at €45.99 each (rent)

templates.dist/fields.php
```diff
@@ -223,3 +223,94 @@ function renderRegionalAddition(string $inputName, array $selectedValue, bool $m
...
+	{
+		$digit = intval(substr(strrchr($str, "."), 1));
+		if (class_exists(NumberFormatter::class)) {
+			$format = new NumberFormatter($locale, NumberFormatter::CURRENCY);
+			if ($digit) {
+				$format->setAttribute(NumberFormatter::MAX_FRACTION_DIGITS, 2);
+			} else {
+				$format->setAttribute(NumberFormatter::MAX_FRACTION_DIGITS, 0);
+			}
+			return str_replace("\xc2\xa0", " ", $format->formatCurrency($str, "EUR"));
```
jmaas-onoffice, line 258: Please use the currency from the estate (field `waehrung`). Each estate can potentially have a different currency.

Also update the other places where the currency is hard-coded to use the estate's currency.

jmaas-onoffice (changes requested):
A few details to tweak, please.

templates.dist/fields.php
tang-hien-egs: sorry, i will fix it as soon !

LongTrong-exe (dismissed):

jmaas-onoffice (approved):
I will test it.

jmaas-onoffice (changes requested):
1. My requirements were a bit confusing, sorry. If there are multiple spaces, they are currently output as an HTML list:

    grafik

    But they should be simply comma-separated like this:

    grafik

   I have put the current requirements into this issue: https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/295

3. The currency of the estate is not respected in the parking lot prices. In the following screenshot, all prices should be in £:

    grafik

3. When there are no parking spaces, the field should not be listed. In this example, the "living area" field is only available for the left estate, while the field is not shown on the right:

    grafik
    
    The same should happen with the "parking" field. In this example, it should not show up for any of the two estates.

dai-eastgate:
For the issue 2: Please change your language to English (UK). When the language is not the same with the currency it will fall back to `€` currency.

jmaas-onoffice:
About the translations:

Concatenating individual translations (like `__("Your city is ", "my-plugin") . $city . __(", and your zip code is ", "my-plugin") . $zipcode`) has the problem that sometimes you need to change the order of the words in different languages. So whenever you have to combine different pieces and make them translatable, it is better to use [this pattern](https://stackoverflow.com/a/60958123) (the comment is important so that the translators know which is which):

```
sprintf(
     /* translators: 1: Name of a city 2: ZIP code */
    __( 'Your city is %1$s, and your zip code is %2$s.', 'my-plugin' ),
    $city,
    $zipcode
);
```

LongTrong-exe (approved):

templates.dist/fields.php
```diff
@@ -269,24 +277,37 @@ function renderParkingLot(array $parkingArray, string $language, string $locale,
...
+				$format->setAttribute( NumberFormatter::MAX_FRACTION_DIGITS, 2 );
 			} else {
-				$format->setAttribute(NumberFormatter::MAX_FRACTION_DIGITS, 0);
-			}
-			if (empty($codeCurrency))
-			{
-				$codeCurrency = "EUR";
+				$format->setAttribute( NumberFormatter::MAX_FRACTION_DIGITS, 0 );
 			}
+			if ( $currency == '?' ) {
```
jmaas-onoffice, line 302: I don't understand why there is so much code around the currency. We should simply get the estate's currency and use that.

jmaas-onoffice (changes requested):
1. The issues 1. and 2. from https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/105#pullrequestreview-1032539627 are still present.
2. There is an extra space before the comma for multiple parking entries:

    grafik

I think it's best to discuss this further in our weekly on Wednesday.

jmaas-onoffice:
I improved the requirements in #295. If you have any questions about them, feel free to ask. :)

jmaas-onoffice (changes requested):
Please implement the requirements as updated in https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/295.

LongTrong-exe (approved):

templates.dist/fields.php
```diff
@@ -248,96 +248,55 @@ function renderRegionalAddition(string $inputName, array $selectedValue, bool $m
...
-				$MarketingType = (' (' . $parking['MarketingType'] . ') ');
+			if ( ! empty( $parking['MarketingType'] ) ) {
+					$MarketingType = ( ' (' . $MarketingType . ')' );
 			}
-			/* translators: 1: Name of parking lot, 2: Price , 3: Each , 4: MarketingType */
-			$element = sprintf( __( '%1$s ' . $preposition . ' %2$s' . '%3$s' . '%4$s', 'onoffice' ), getParkingName($key, $parking['Count']), formatPriceParking($parking['Price'], $language, $locale, $codeCurrency, $currency), $pluralPaking, $MarketingType );
-			
-			array_push($messages, $element);
+			/* translators: 1: Name of parking lot, 2: Preposition, 3: Price, 4: Each, 5: MarketingType */
+			$element = sprintf( __( '%1$s ' . '%2$s' . ' %3$s' . '%4$s' . '%5$s', 'onoffice' ), $parkingName,
```
jmaas-onoffice, line 277: This will not be translatable (see #317). String concatenations are not recognized and you are using variables where you could just place it into the translation string.

There is a complication because we need to translate the parking name differently, depending on whether it is singular or plural.

One possibility:

```php
if ($key == "garage") {
  $countAndName = _n("%s garage", "%s garages", $count);
} else if ($key == "multiStoryGarage") {
  $countAndName = _n("%s multi-story garage", "%s multi-story garages", $count);
} else if ...

$marketingTypeOutput = "";

$marketingType = $parking['MarketingType'];
if ( ! empty($marketingType) ) {
  if ($marketingType == "rent") {
    $marketingTypeOutput = __(" (rent)", 'onoffice-for-wp-websites');
  } else if ($marketingType == "buy") {
    $marketingTypeOutput = __(" (purchase)", 'onoffice-for-wp-websites');
  } else {
    // Throw an error about unrecognized marketing type.
  }
}

$price = formatPriceParking( $parking['Price'], $language, $locale, $codeCurrency, $currency );

/* translators: 1: Count and name of parking lot, 2: Price, 3: Marketing type */
$element = sprintf( _n( '%1$s at %2$s%3$s', '%1$s at %2$s each%3$s', $count, 'onoffice-for-wp-websites' ), $countAndName, $price, $marketingTypeOutput);

File: templates.dist/fields.php [hunk 22]
```diff
@@ -248,96 +248,55 @@ function renderRegionalAddition(string $inputName, array $selectedValue, bool $m
...
 			if ( $parking['Count'] != 1 ) {
 				$pluralPaking = " each";
 			}
-			if ($codeCurrency == "EUR" && $locale == "de_DE" ) {
-				$preposition = "à";
-				$pluralPaking = "";
+
+			$parkingName = getParkingName( $key, $parking['Count'] );
+
+			if ( $codeCurrency == "EUR" && $locale == "de_DE" ) {
```
jmaas-onoffice, line 267: We should translate using the gettext functions (e.g. `__()`). This condition also has the problem that it would not display parking lots in German, just because the estate has a non-euro currency. But the website is in German so it should still be output in German!

jmaas-onoffice (changes requested):
It looks much cleaner already!

Just a few things I found.

jmaas-onoffice (approved):

jmaas-onoffice:
We noticed that we would have to change all the templates for existing users to make the parking lots work again. So I will investigate if there is a way so that existing user have the field work automatically.

Unfortunately, currently I'm quite busy, so it might take a while. But from my tests it looks like we implemented all the requirements now, so this is just something that was missing in the requirements. ;)

jmaas-onoffice:
Finally! 🎉 Thanks everyone, I tested it and it looks good! :)

PR #112, opened 2021-09-07
P#51845 Fix error for MonsterInsights setup

hungnc89:


hungnc89:
I think the id field of current_screen is null because MonsterInsights setup wizard page doesn't use admin page template of wp, so current_screen's fields were not set

jayay (approved):

PR #113, opened 2021-09-07
P#53033(Task) - Fix error save form with dummy_key

hungnc89:


js/ajax_settings.js
```diff
@@ -63,7 +63,7 @@ onOffice.ajaxSaver = function(outerDiv) {
 		this._outerDiv.find('.onoffice-input:not([data-onoffice-ignore=true])').each(function(i, elem) {
 			var inputNameFull = $(elem).attr('name');
 			var inputName = inputNameFull;
-			if (inputName === "dummy_key") {
+			if (inputName === "dummy_key" || inputName.includes("dummy_key")) {
```
jayay, line 66: What's the name, if it just _includes_ "dummy_key"?

hungnc89: Hi, @jayay. I checked and noticed that `generateClone` in `onoffice-custom-form-label.js` created a new dummy_key type field. For example customlabel-lang[dummy_key][native]
I'm not sure if it adds any other formatting to the dummy_key field. So I use `includes`

jayay (approved):

PR #114, opened 2021-09-10
P#51847 Compatible default template and show template names

hungnc89:


plugin/Renderer/InputFieldTemplateListRenderer.php
```diff
@@ -0,0 +1,162 @@
...
+		foreach ($this->getValue() as $templateValue) {
+			$templateList = $templateValue['path'];
+			if (count($this->getValue()) > 1) {
+				if (in_array($this->_checkedValue, $templateList) ||
+					array_key_exists($this->_checkedValue, $templateList)) {
+					echo '<details open>';
+				} else {
+					echo '<details>';
+				}
+				echo '<summary>' . $templateValue['title'] . '</summary>';
```
jayay, line 85: `esc_html()` would be highly appreciated here

plugin/Renderer/InputFieldTemplateListRenderer.php
```diff
@@ -0,0 +1,162 @@
...
+				}
+				$inputId = 'label' . $this->getGuiId() . 'b' . $key;
+				echo '<input type="' . esc_html($this->getType()) . '" name="' . esc_html($this->getName())
+					. '" value="' . esc_html($key) . '"'
+					. ($checked ? ' checked="checked" ' : '')
+					. $this->renderAdditionalAttributes()
+					. ' id="' . esc_html($inputId) . '">'
+					. '<label for="' . esc_html($inputId) . '">' . esc_html($label) . '</label><br>';
+			}
+			echo "<p>(in the folder " . $templateValue['folder'] . ")</p>";
```
jayay, line 101: same here

plugin/Model/FormModelBuilder/FormModelBuilder.php
```diff
@@ -125,16 +125,29 @@ protected function readTemplatePaths($directory, $pattern = '*')
 		$templatesAll = array_merge($templateGlobFiles, $templateLocalFiles, $templateThemeFiles);
 		$templates = array();
 
+		$plugin_name = basename(plugin_dir_path(ONOFFICE_PLUGIN_DIR . '/index.php'));
 		foreach ($templatesAll as $value) {
+			$file = substr(strrchr($value, "/"), 1);
 			if(strpos($value, 'themes') !== false) {
 				$value = __String::getNew($value)->replace(get_template_directory().'/', '');
-				$templates[$value] = $value;
```
jayay, line 131: Does it work to hardcode `$template[1]`, `$template[2]` and `$template[3]` here?

jayay (commented):
A couple of questions came up

plugin/Model/FormModelBuilder/FormModelBuilder.php
hungnc89: Keys 1, 2, 3 are the order of the template folders when displayed. Maybe I'll try using an array to store order of the template folders instead of hardcode

plugin/Model/FormModelBuilder/FormModelBuilder.php
```diff
@@ -50,6 +50,13 @@ abstract class FormModelBuilder
 	/** @var Fieldnames */
 	private $_pFieldnames = null;
 
+	/** @var int[] */
+	private $_orderOfTemplatesFolder = [
```
jayay, line 54: Please use `const` for data that is not supposed to be changed.

plugin/Model/FormModelBuilder/FormModelBuilder.php
```diff
@@ -130,19 +137,22 @@ protected function readTemplatePaths($directory, $pattern = '*')
...
+				$templates[$folderOrder]['title'] = 'Personalized (Theme)';
+				$templates[$folderOrder]['folder'] = '/onoffice-theme/templates/' . $directory . '/';
 			}else{
 				$value = __String::getNew($value)->replace(plugin_dir_path(ONOFFICE_PLUGIN_DIR), '');
 				if (strpos($value, 'onoffice-personalized') !== false) {
-					$templates[2]['path'][$value] = $file;
-					$templates[2]['title'] = 'Personalized (Plugin)';
-					$templates[2]['folder'] = 'onoffice-personalized/templates/' . $directory . '/';
+					$folderOrder = $this->getOrderOfTemplateFolder('plugin');
+					$templates[$folderOrder]['path'][$value] = $file;
```
jayay, line 148: one `$templates[$folderOrder]` can only have one template?

jayay (commented):
Thanks for the fixes. I wrote down more questions.

Do you think it'd be possible to write more tests?

plugin/Model/FormModelBuilder/FormModelBuilder.php
hungnc89: @jayay I think not, `$templates` is an array to store templates folder, so `$templates[$folderOrder]` will include information about that folder and also its templates, I think it's fine to use this way, if you have any good method more, please describe to me.

jayay: Guess I didn't understand `$templates[$folderOrder]['path'][$value] = $file`.

Do you think a data model would be easier to deal with here?

hungnc89: I think it doesn't make much difference

nglelinh:
@jayay could you please help to continue to review this PR ?

plugin/Model/FormModelBuilder/FormModelBuilder.php
```diff
@@ -123,19 +129,35 @@ protected function readTemplatePaths($directory, $pattern = '*')
...
-		$templates = array();
-
-		foreach ($templatesAll as $value) {
-			if(strpos($value, 'themes') !== false) {
-				$value = __String::getNew($value)->replace(get_template_directory().'/', '');
-				$templates[$value] = $value;
-			}else{
-				$value = __String::getNew($value)->replace(plugin_dir_path(ONOFFICE_PLUGIN_DIR), '');
-				$templates[$value] = $value;
+		$templateFolderData = array();
```
jayay, line 132: This is a very central place to put this rather special code. It'd be best if it was somewhere where it actually belongs and if it was tested.

jayay (approved):

jayay (changes requested):
Sorry, I pressed the wrong button. The new code should be testable and moved away from `FormModelBuilder` because that's too central.

tests/TestClassTemplateCall.php
```diff
@@ -55,6 +55,14 @@ class TestClassTemplateCall
 
 	/** @var array */
 	private $_expectationTemplatePath = [
+		1 => [
+			'path' => [
+				'...\wordpress\wp-content\plugins/onoffice-themes/templates/address/SearchFormAddress.php' => "SearchFormAddress.php",
```
jayay, line 60: Where's that back-slash/forward-slash mix coming from?

hungnc89: @jayay It comes from this function
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/a789d6b1f81acffea9cdbec994a9325e3489fd34/plugin/Model/FormModelBuilder/FormModelBuilder.php#L116
Glob function will return the path separated by backslash

hungnc89: I don't think this has any effect on this function

nglelinh:
@jayay could you please help to review again?

jayay (approved):

michaelarnold-dev:
@hungnc89 Testing feedback by Robert Igelmund :

Requirement: For each template, the file name is displayed. The folder's path is given in the style of setting descriptions (for example for the setting "Settings > General > Tagline"). 

The settings descriptions are black, not gray as requested.The other differences in the GUI from the screenshot can remain as they are.

Could you please fix that?

plugin/Renderer/InputFieldTemplateListRenderer.php
```diff
@@ -98,7 +98,7 @@ public function render()
 					. ' id="' . esc_html($inputId) . '">'
 					. '<label for="' . esc_html($inputId) . '">' . esc_html($label) . '</label><br>';
 			}
-			echo "<p>". esc_html("(in the folder " . $templateValue['folder'] . ")") ."</p>";
+			echo "<p class='oo-template-folder-path'>". esc_html("(in the folder " . $templateValue['folder'] . ")") ."</p>";
```
jayay: Translation is missing here

jayay (changes requested):

jayay (approved):

jayay:
@LongTrong-exe would you please update this branch? This lead to re-declarations so I reverted my attempt.

jayay (approved):
Thank you!

michaelarnold-dev:
@LongTrong-exe The following video shows that it can not be saved at the end
"There was a problem with saving the view" Could you please analyze?

https://user-images.githubusercontent.com/77264788/160079850-82fa7e47-73ae-4c1e-bb72-ae18b3be4385.mp4

LongTrong-exe:
@michaelarnold-dev About 1), I think you should try pressing "ctrl + shift + R" to reload page and delete cache. New style changes will be applied.

LongTrong-exe:
@michaelarnold-dev This is my test video:

https://user-images.githubusercontent.com/101325858/160085114-7d45989f-b70f-4260-bee6-d409b8e6bffa.mp4

michaelarnold-dev:
@LongTrong-exe 1) is fixed by clearing the cache. But I wonder why I have to clear the cache manually, it happens in some projects but it never says so. Is there no way to automate this? And are the customers not affected by this? 2) and 3) still exist.

LongTrong-exe:
@michaelarnold-dev I think bug 2, 3 happen when u test on different versions with same test data.
Make sure every time you test on a new version, re-create data from master(Version 2.0).

jmaas-onoffice (changes requested):
Can you revert this change? While this is a clever way to make it work, this is not the most obvious place to do this.

For now, we will manually clear the cache while testing.

jmaas-onoffice (approved):
Thank you! I will coordinate the testing now. ;)

jmaas-onoffice (changes requested):
I think the requirement 2.2) wasn't clear. Let me explain a bit better:

Usually the user will choose to put the personalized tempalte in the plugin folder _or_ the theme folder, but rarely both. We think it is confusing for the user to read "Personalized (Plugin)".

Therefore, if the user only has personalized templates in one of those places, we want to just label it "Personalized" (without "(Plugin)" or "(Theme)". Only if the user has template files in both locations to we need to distinguish them as "Personalized (Plugin)" and "Personalized (Theme)".

Example:
1. User only has personalized templates in "/wp-content/plugins/onoffice-personalized":
  - Personalized
  - Default
2. User only has personalized templates in "/wp-content/themes/onoffice-theme":
  - Personalized
  - Default
3. User has personalized templates in "/wp-content/plugins/onoffice-personalized" and  "/wp-content/themes/onoffice-theme":
  - Personalized (Theme)
  - Personalized (Plugin)
  - Default

I hope this is a bit clearer. If there are any questions, feel free to ask here.

Please change the code so that the "Personalized" label is as simple as possible, as described above.

plugin/Template/TemplateCall.php
```diff
@@ -94,39 +94,52 @@ public function loadTemplates()
...
+			$templateFolderData[ $folderOrder ]['folder'] = $shortPath;
+		}
+		if ( $hasTemplateFolder !== $hasPluginFolder ) {
+			foreach ( $templateFolderData as $key => $templateData ) {
+				if ( strpos( $templateData['title'], "Personalized" ) !== false ) {
+					$templateFolderData[ $key ]['title'] = "Personalized";
+				}
+			}
 		}
+		ksort( $templateFolderData );
```
jmaas-onoffice, line 141: Can this be simplified?

For example, you [merge](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/114/files/639c4d0e062eb263c57b65ae0956266d087da159..c8ee5df2468b3de82ced360529efbdf07e053fbc#diff-b77363609ca4a4f37d1ecd149c1d034d6ff43a880bd44945099368e1762a72d9R123) all the paths together before passing them in as `$templatesAll`, but then in this function you separate them again.

So it seems that maybe you can keep the template locations separated so that you avoid the need to separate them in this function.

I hope this comment is understandable. If you have questions, feel free to ask here. :)

jmaas-onoffice (changes requested):
I made a mistake with my previous comment, sorry. (There is also one idea for simplifying the code.)

plugin/Template/TemplateCall.php
LongTrong-exe: Oh, nice idea, I'll try it.

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
I noticed a small issue:

On a new install of the plugin, the detail view and similar estates of course already exist. But they do not have a template set yet until you save them for the first time.

This means that you can cause an "Invalid template path" error as follows:
1. Make a fresh install of the plugin. (Make sure that the database contains no entries for the plugin, especially the "options" table.)
2. Add the API credentials.
3. Make a new page where you add the detail view shortcode.
4. Visit the detail page for some estate (add the estate ID to the URL).
5. See the error:

        [onOffice-Plugin] An error occured:
        RuntimeException: Invalid template path in /www/htdocs/w01bb2af/testinstanz-jmaa.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Template.php:140

Could you please make it so that the estate view and similar estates are set to the suitable template when installed?

LongTrong-exe:
Uploading 2022-04-29 14-40-55.mp4…

jmaas-onoffice (approved):

PR #118, opened 2021-09-28
P#54045 Meta title and description for detail view

vu-vuong-esg:
P#54045 [WP-Plugin] Meta title and description für detail view


jayay:
It'd be nice if you could add some tests and if the user interface strings could be translated.

vu-vuong-esg:
> It'd be nice if you could add some tests and if the user interface strings could be translated.

Translation and unit were added. Please have a look.

vu-vuong-esg:
> It'd be nice if you could add some tests and if the user interface strings could be translated.

Translation and unit test were added. But the coverage still decreased -0.02%

michaelarnold-dev:
@tung-le-esg  We added a note to 4.2) that the custom fields should only be available for detail pages, because they cannot be set correctly on any other page."

jayay (changes requested):
A loop would have been easier to understand and be more clean than the ternary operator without an assignment to a variable.

This should be its own class, because EstateList is hard to test.

plugin/Controller/AdminViewController.php
```diff
@@ -397,4 +397,17 @@ public function displayAPIError()
 			printf('<div class="%1$s"><p>%2$s</p></div>', esc_attr($class), $message);
 		}
 	}
+
+	public function general_admin_notice_SEO(){
+		$pWPOptionWrapper = new WPOptionWrapperDefault();
+		if ( count(array_intersect(["wordpress-seo/wp-seo.php", "seo-by-rank-math/rank-math.php", "wpseo/wpseo.php"], get_option("active_plugins"))) > 0 && $pWPOptionWrapper->getOption('onoffice-settings-title-and-description') == 0) {
+			echo '<div class="notice notice-warning is-dismissible">
+						<p> '.__('The onOffice plugin has detected an active SEO plugin: Yoast SEO.'
```
jayay, line 405: `esc_html__` should be used here

plugin/EstateList.php
```diff
@@ -397,9 +402,46 @@ public function estateIterator($modifier = EstateViewFieldModifierTypes::MODIFIE
...
 			$recordModified['vermarktungsstatus'] = $pEstateStatusLabel->getLabel($recordRaw);
 		}
+		if ($this->_pWPOptionWrapper->getOption('onoffice-settings-title-and-description') == 0)
+		{
+			$recordModified['onoffice-settings-title-and-description'] = true;
+		}
+		$lang = Language::getDefault();
+		if ($this->_pWPOptionWrapper->getOption('onoffice-settings-title-and-description') == 1)
+		{
+			$listCustomField = [
```
jayay, line 412: Thank you for refactoring this. Sadly, EstateList grows and grows but can hardly be tested. That's why this new block should be in its own class.

plugin/Gui/AdminPageApiSettings.php
```diff
@@ -166,18 +166,49 @@ private function addFormModelGoogleMapsKey()
...
+			$labelTitleAndDescription, InputModelOption::SETTING_TYPE_NUMBER);
+		$pInputModeTitleAndDescription->setHtmlType(InputModelOption::HTML_TYPE_RADIO);
+		$pInputModeTitleAndDescription->setValuesAvailable([
+			__('Fill out', 'onoffice-for-wp-websites'),
+			__('Do not modify', 'onoffice-for-wp-websites'),
+		]);
+		$pInputModeTitleAndDescription->setValue(get_option($pInputModeTitleAndDescription->getIdentifier()));
+		$pInputModeTitleAndDescription->setDescriptionRadioTextHtmls([
+			add_action('admin_notices', [$pAdminViewController, 'general_admin_notice_SEO']).__('This plugin will fill out the title and description with the information from the estate that is shown. This option is recommended if you are not using a SEO plugin.', 'onoffice-for-wp-websites'),
+			__("This plugin will not modify the title and description. This enables other plugins to manage those tags.",'onoffice-for-wp-websites').
```
jayay, line 188: the `__()` calls should be turned into `esc_html__()` function calls.

templates.dist/estate/default_detail.php
```diff
@@ -33,6 +33,17 @@
 	<?php
 	$pEstates->resetEstateIterator();
 	while ( $currentEstate = $pEstates->estateIterator() ) { ?>
+		<?php
+		if($currentEstate["onoffice-settings-title-and-description"])
```
jayay, line 37: `<head>` is usually done by the WP theme.  I think this whole thing needs to run a lot sooner, similar to [EstateViewDocumentTitleBuilder](https://github.com/onOfficeGmbH/oo-wp-plugin/blob/057cca241c8efe32bea6514ddf3e46f764e1122e/plugin.php#L131-L133).

jayay (changes requested):
Please see my comments. I don't think `EstateList` is the right place for this. There's another feature that builds `<title>` tags, please see my comment. It should be done more similar to that.

vu-vuong-esg:
> Please see my comments. I don't think `EstateList` is the right place for this. There's another feature that builds `<title>` tags, please see my comment. It should be done more similar to that.

Yes. I will check it.

plugin/Gui/AdminPageApiSettings.php
```diff
@@ -166,18 +166,51 @@ private function addFormModelGoogleMapsKey()
...
+		$labelTitleAndDescription = __('Title and description', 'onoffice-for-wp-websites');
+		$pInputModeTitleAndDescription = new InputModelOption('onoffice-settings', 'title-and-description',
+			$labelTitleAndDescription, InputModelOption::SETTING_TYPE_NUMBER);
+		$pInputModeTitleAndDescription->setHtmlType(InputModelOption::HTML_TYPE_RADIO);
+		$pInputModeTitleAndDescription->setValuesAvailable([
+			__('Fill out', 'onoffice-for-wp-websites'),
+			__('Do not modify', 'onoffice-for-wp-websites'),
+		]);
+		$pInputModeTitleAndDescription->setValue(get_option($pInputModeTitleAndDescription->getIdentifier()));
+		$pInputModeTitleAndDescription->setDescriptionRadioTextHtmls([
```
jayay, line 188: We will regret passing HTML into the setter. How about moving the formatting into the renderer? Would that work?

plugin/Model/InputModelBase.php
```diff
@@ -247,4 +250,12 @@ public function getDescriptionTextHTML()
     public function setDescriptionTextHTML(string $textHTML)
     	{ $this->_descriptionTextHTML = $textHTML; }
 
+
+	/** @return string */
+	public function getDescriptionRadioTextHtmls()
```
jayay, line 255: This is a little too specific for a _base class_.

jayay (changes requested):

plugin/EstateList.php
```diff
@@ -397,9 +402,21 @@ public function estateIterator($modifier = EstateViewFieldModifierTypes::MODIFIE
 			$pEstateStatusLabel = $this->_pEnvironment->getEstateStatusLabel();
 			$recordModified['vermarktungsstatus'] = $pEstateStatusLabel->getLabel($recordRaw);
 		}
-
+		if ($this->_pWPOptionWrapper->getOption('onoffice-settings-title-and-description') == 0)
+		{
+			add_action('wp_head',function ($recordModified){
+			?>
+				<title><?php echo $recordModified["objekttitel"]; ?></title>
```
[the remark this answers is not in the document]
jayay: Have you seen https://codex.wordpress.org/Meta_Tags_in_WordPress#Using_Meta_Tag_Plugins? I think that is a lot more what we're looking for.

hungnc89:

jayay:
Have you seen [this comment](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/118/files?file-filters%5B%5D=.php&show-viewed-files=true#r789753409) about modifying meta parameters?

hungnc89:

plugin/EstateListHandle.php
```diff
@@ -0,0 +1,53 @@
...
+										'objektart' => apply_filters('get_post_metadata','onoffice_objektart'),
+										'vermarktungsart' => apply_filters('get_post_metadata','onoffice_vermarktungsart'),
+										'Id' => apply_filters('get_post_metadata','onoffice_datensatznr')
+									]
+						];
+	}
+	public function handleRecord($recordModified)
+	{
+		foreach ($recordModified as $key => $record) {
+			$lang = Language::getDefault();
```
jayay, line 35: I think both English and German names always need to be accessible, otherwise things can break if two different users are using WP in different languages

plugin/EstateListHandle.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace onOffice\WPlugin;
+
+class EstateListHandle
+{
+
+	private $_listCustomField = array();
```
jayay, line 8: Would it be possible to eliminate `$_listCustomField ` to make the class stateless?

jayay (changes requested):
Almost there! There's a small things that I would like to be changed

plugin/EstateList.php
```diff
@@ -409,6 +409,19 @@ public function estateIterator($modifier = EstateViewFieldModifierTypes::MODIFIE
...
+		{
+			add_filter('document_title_parts', function() use ($recordModified) {
+				if (isset($recordModified["objekttitel"]))
+				{
+					return [$recordModified["objekttitel"]];
+				}
+				return [];
+			}, 10, 2);
+			add_action( 'wp_head', function () use($recordModified) {
+				echo '<meta name="description" content="' . esc_attr($recordModified["objektbeschreibung"] ?? null) . '" />';
```
jayay, line 422: Isn't it way too late to trigger `wp_head` at this point?

js/onoffice-setting-api.js
```diff
@@ -0,0 +1,7 @@
+jQuery(document).ready(function($) {
+    var url = window.location.href;
+    if ( url.search('onoffice-settings') != -1 && document.querySelector('.notice-warning') !== null )
```
jayay, line 3: let's say I own the domain onoffice-settings.com...

jayay: This check should be moved into the place where the notification gets displayed

plugin/EstateList.php
jayay: Still unsolved

tang-hien-egs: @jayay , i only know hook wp-head change tag meta description and i add priority with PHP_INT_MAX for this hook to it, and you review PR and suggest comment to finish this task .

templates.dist/estate/default_detail.php
```diff
@@ -32,7 +35,17 @@
 <div class="oo-detailview">
 	<?php
 	$pEstates->resetEstateIterator();
-	while ( $currentEstate = $pEstates->estateIterator() ) { ?>
+    $pContainerBuilder = new ContainerBuilder;
```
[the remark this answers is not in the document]
tang-hien-egs: > Please do not expose the container to the template. `get_option()` is totally legit here

i fixed it with use get_option().

plugin/EstateListHandle.php
```diff
@@ -0,0 +1,81 @@
...
+	/**
+	 * @var array[]
+	 */
+	private $_listCustomField;
+	public function __construct()
+	{
+		$pContainerBuilder = new ContainerBuilder;
+		$pContainerBuilder->addDefinitions(ONOFFICE_DI_CONFIG_PATH);
+		$pContainer = $pContainerBuilder->build();
+		$pEnvironment = new EstateListEnvironmentDefault($pContainer);
```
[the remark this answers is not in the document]
tang-hien-egs: > `EstateListEnvironmentDefault` must only be used by `EstateList`

i fixed it , and use DataDetailViewHandler

tang-hien-egs:
@jayay , can you check PR help me ?

jmaas-onoffice (changes requested):
I've reviewed together with Jakob. There are still a few changes we would like to be made.

plugin/EstateList.php
```diff
@@ -405,9 +413,20 @@ public function estateIterator($modifier = EstateViewFieldModifierTypes::MODIFIE
...
+			add_filter('pre_get_document_title', function($title_parts_array) use ($recordModified) {
+				if (isset($recordModified["objekttitel"]))
+				{
+					$title_parts_array = $recordModified["objekttitel"];
+				}
+				return $title_parts_array;
+			},999,1);
+			add_action( 'wp_head', function () use($recordModified) {
+				echo '<meta name="description" content="' . esc_attr($recordModified["objektbeschreibung"] ?? null) . '" />';
+			},PHP_INT_MAX);
```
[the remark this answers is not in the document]
vu-vuong-esg: about priority of hook, Lower numbers correspond with earlier execution, and functions with the same priority are executed in the order in which they were added to the action, and i test priority of hook with add 2 hook with priority other, link video : “https://files.fm/f/w7nwkgtxp ” , you suggest and comment when i add priority is 1

plugin/EstateList.php
vu-vuong-esg: > Ok, since this seems to work (even though we don't understand why), we will use this `wp_head` hook and test it.

@jmaas-onoffice when display tag meta in <head>, we create 1 function hook in action wp_head, The wp_head action hook is triggered within the <head></head> section of the theme’s header.php template by the wp_head() function.

plugin/EstateListHandle.php
vu-vuong-esg: > This still needs to changed: Don't check which language is active. Just add all the fields of both languages.

@jmaas-onoffice field ‘objekttitel’ will contain value both custom field ‘onoffice_titel’ and ‘onoffice_title’. Link video demo : https://files.fm/f/yswfnrkg6. is this your mean ?

plugin/EstateList.php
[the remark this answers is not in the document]
jmaas-onoffice: I'm not sure what your question is or what you are showing in the video.

The problem with PHP_INT_MAX is that no one can enqueue something after that, and there is no reason to do that here. That's why you should choose a more reasonable number like 20, but you need to check if that is high enough.

templates.dist/estate/default_detail.php
```diff
@@ -32,7 +35,13 @@
 <div class="oo-detailview">
 	<?php
 	$pEstates->resetEstateIterator();
-	while ( $currentEstate = $pEstates->estateIterator() ) { ?>
+	while ( $currentEstate = $pEstates->estateIterator() ) {
+		if (get_option('onoffice-settings-title-and-description') == 1)
+		{
+			$estateListHandle = new EstateListHandle();
+			$currentEstate = $estateListHandle->handleRecord($currentEstate);
```
jmaas-onoffice, line 42: Could you explain to me what this does? I don't understand what problem it solves.

templates.dist/estate/default_detail.php
```diff
@@ -17,10 +17,13 @@
  *    along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
  */
-
-use onOffice\WPlugin\EstateDetail;
-
-/**
+	
+	use DI\ContainerBuilder;
```
jmaas-onoffice, line 21: Please clean up the unused imports.

plugin/EstateList.php
jmaas-onoffice: We just thought that since we are in the template, we have already sent the `<head>`. So we thought that we couldn't add anything to the head at this point.

But if you say this works, then we'll just test it. 👍

plugin/EstateListHandle.php
jmaas-onoffice: Unfortunately it seems there was a misunderstanding. The video does not look like it is what we wanted.

I will prepare a description of what problem we want to solve and present this in the weekly with Linh.

jmaas-onoffice (commented):
There seems to be a misunderstanding with the requirements. I will improve the requirements and discuss them with Linh in the next weekly.

plugin/EstateListHandle.php
vu-vuong-esg: > Unfortunately it seems there was a misunderstanding. The video does not look like it is what we wanted.
> 
> I will prepare a description of what problem we want to solve and present this in the weekly with Linh.

i don't understand your mean, please demo detail your mean with video !

templates.dist/estate/default_detail.php
vu-vuong-esg: i check when user selected option "Do not modify", i will change value of field "objekttitel" =  value of custom field " `onoffice_titel` / `onoffice_title`"

vu-vuong-esg:
@j i have record video demo your solution to resolve PR , please watch to video , like link : https://files.fm/f/pqhhh9wah , if it not your requirement , please demo requirement with video help me

jmaas-onoffice (changes requested):
Linh took a video of my explanation. Here's a brief summary.

When you use an SEO plugin like [Yoast](https://yoast.com/), it will set the `<title>` and `<meta name="description">` for each page. For the detail page, the problem is that in WordPress the title is for example "Detail". So every estate detail page will have the same title "Detail".

We want to take the estate's title (and other fields) from enterprise and give them to Yoast. One way of doing that is to use Yoast's ability to [use custom fields](https://yoast.com/help/list-available-snippet-variables-yoast-seo/#custom-fields). WordPress actually gives us a [hook to return values for custom fields](https://developer.wordpress.org/reference/hooks/get_meta_type_metadata/). So if we use that hook and for e.g. `onoffice_title` we return the actual title of the current estate, we can use that custom in Yoast like this:

grafik

So we want to pass information from enterprise to SEO plugins like Yoast, and a way to do that is hacking custom fields.

One additional thing: To make it easier for German users, we want to not only use English names for the custom fields, but also German names. So English users should be able to use `onoffice_description` and German users `onoffice_beschreibung` to use the estate's description in the SEO plugin. But there is no need to limit which names are available by language. Even if someone has set their WordPress to English, they should still be able to use `onoffice_beschreibung`.

If there are questions, maybe Linh can help. Otherwise feel free to ask here! :)

LongTrong-exe (dismissed):

LongTrong-exe (dismissed):

tang-hien-egs:
@jmaas-onoffice 
For requirement : “until the user dismisses the warning by clicking on the "x" , 
I want clean it : 

when I select “Fill out” and plugin “Yoast SEO” active, warning will display. I click “x” then warming will temporarily closed (refesh F5 still display) or closed forever ?

https://user-images.githubusercontent.com/95266110/168543657-fb5aa4f8-b19f-40ae-8d46-4f324b4b56a0.mp4

jmaas-onoffice:
@tang-hien-egs Thanks for the question! When the user clicks the "x", it should stay closed forever.

tang-hien-egs:
> @tang-hien-egs Thanks for the question! When the user clicks the "x", it should stay closed forever.

yes, thanks you , i will update it !

hungnc89:
- Tester tested
- Video: 

https://user-images.githubusercontent.com/72601287/168583526-cd7a1503-601a-4235-a034-5c16849c3438.mp4

tang-hien-egs:
@jmaas-onoffice , i fixed PR, please review it help me !

tang-hien-egs:
> On the detail page at 1:40, the warning was displayed twice. It should only be displayed one time. (Note that the second breadcrumbs will be removed in another PR: [#234 (comment)](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/234#issuecomment-1122171173))
> 
> Apart from that, the video looks good. When you've fixed this small issue, please request another review and I will test it. :)

 cause issue fixed in PR #234, and i tested local is oki, should we update it at here or wait merge wit

https://user-images.githubusercontent.com/95266110/168728116-b2448bc1-db84-4252-949f-3050f7805860.mp4

h PR #234 ?

tang-hien-egs:
> I think #234 will still take a while. Do you think it's easy to fix the same thing here (and the double warning)? Or will there be a difficult merge conflict in #234?

it's easy to fix the same at here ! So i will update it at here to finish PR !

tang-hien-egs:
@jmaas-onoffice , i updated , please review help me !

jmaas-onoffice (dismissed):

plugin/Controller/AdminViewController.php
```diff
@@ -417,4 +422,27 @@ public function getField()
...
+		$nameOnofficeSetting = esc_html__('onOffice plugin settings','onoffice-for-wp-websites');
+		$pluginOnofficeSetting = sprintf("<a href='%s' target='_blank' rel='noopener'>%s</a>", $urlOnofficeSetting,$nameOnofficeSetting);
+		if (count(array_intersect(["wordpress-seo/wp-seo.php", "seo-by-rank-math/rank-math.php", "wpseo/wpseo.php"], get_option("active_plugins"))) > 0) {
+			if (get_option('onoffice-click-button-close-action') == 0
+				&& get_current_screen()->id !== 'onoffice_page_onoffice-settings'
+				&& get_option('onoffice-settings-title-and-description') == 0) {
+				echo "<div class='notice notice-warning active-plugin-seo is-dismissible'>
+						<p> " . esc_html__('The onOffice plugin has detected an active SEO plugin: Yoast SEO.', 'onoffice-for-wp-websites') . "
+							" . esc_html__('You currently have configured the onOffice plugin to fill out the title and description of the detail page, which can lead to conflicts with the SEO plugin.', 'onoffice-for-wp-websites') . "<br>
+							" . esc_html__('We recommend that you go to the', 'onoffice-for-wp-websites') . "
```
jmaas-onoffice, line 437: There are a few improvements I would like to make.

1. It is difficult to translate each sentence separately. Additionally, it is hard to control the placement of the `$pluginOnofficeSetting`. It would be better to put it all in one `esc_html__` and to use `sprintf` to place the `$pluginOnofficeSetting`. That way a translator can place it where they need to.

    Example:
    ```
    sprintf(esc_html__('We recommend that you go to the %s and configure the onOffice plugin to not modify the title and description. This allows you to manage the title and description with your active SEO plugin.', 'onoffice-for-wp-websites'), $pluginOnofficeSetting);
    ```

    Since we need to handle line breaks (and in the other comment you can see that it would be nice to handle lists as well), consider using a Markdown parser like https://github.com/erusev/parsedown. First, translate the text, then put it through the parser. For example:

    ```
    echo Parsedown::instance()->text(
      sprintf(
        esc_html__('We recommend that you go to the %s and configure the onOffice plugin to not modify the title and description. This allows you to manage the title and description with your active SEO plugin.', 'onoffice-for-wp-websites'),
        $pluginOnofficeSetting
      )
    );
    ```

2. The warning should identify the conflicting plugin. Currently it always says "Yoast SEO".

    The desired display names are:
    - `wordpress-seo/wp-seo.php` -> "Yoast SEO"
    - `seo-by-rank-math/rank-math.php` -> "Rank Math SEO"
    - `wpseo/wpseo.php` -> "wpSEO"

plugin/Gui/AdminPageApiSettings.php
```diff
@@ -172,18 +175,69 @@ private function addFormModelGoogleMapsKey()
...
+				.' These custom fields can be used in SEO plugins to fill out the title and description with the information of the currently shown estate.'
+				." For information on how to use custom fields consult you SEO plugin's documentation.",'onoffice-for-wp-websites').'<br>'.
+			esc_html__('These custom fields are only available in the detail view and on no other page.','onoffice-for-wp-websites').'<br>
+					   - '. esc_html__('Title (onoffice_title)','onoffice-for-wp-websites').'<br>
+					   - '. esc_html__('Description (onoffice_description)','onoffice-for-wp-websites').'<br>
+					   - '. esc_html__('Place (onoffice_place)','onoffice-for-wp-websites').'<br>
+					   - '. esc_html__('Postal code (onoffice_postal_code)','onoffice-for-wp-websites').'<br>
+					   - '. esc_html__('Property class (onoffice_property_class)','onoffice-for-wp-websites').'<br>
+					   - '. esc_html__('Marketing method (onoffice_marketing_method)','onoffice-for-wp-websites').'<br>
+					   - '. esc_html__('Data Record Ref No. (onoffice_id)','onoffice-for-wp-websites').'<br>
```
jmaas-onoffice, line 195: Similar to before, the block after the `summary` can be grouped as a single unit to make translation easier. The list should also be a proper HTML list, but if we can use the Markdown approach explained in the other comment, that would be easy.

jmaas-onoffice (changes requested):
I noticed a typo and forgot to update the test, but I have a few improvements I would like to make regarding the translations.

It is sometimes difficult to decide what can be translated together, but in this case it would make the translation much easier.

tang-hien-egs:
@jmaas-onoffice , i still fix issue in PR .

jmaas-onoffice:
@tang-hien-egs I'm not sure what you mean. Do you need anything from me? :)

tang-hien-egs:
> @tang-hien-egs I'm not sure what you mean. Do you need anything from me? :)

No, i clean your comment yesterday. Today, i will fix it !

tang-hien-egs:
@jmaas-onoffice , i fixed in PR, please review it help me !

plugin/Controller/AdminViewController.php
```diff
@@ -427,19 +428,29 @@ public function generalAdminNoticeSEO() {
...
-							" . esc_html__('We recommend that you go to the', 'onoffice-for-wp-websites') . "
-							" . $pluginOnofficeSetting . "
-							" . esc_html__('and configure the onOffice plugin to not modify the title and description.', 'onoffice-for-wp-websites') . "
-							" . esc_html__('This allows you to manage the title and description with your active SEO plugin.', 'onoffice-for-wp-websites')
-						. "</p>
-					</div>";
+				$class = 'notice notice-warning active-plugin-seo is-dismissible';
+				$message = sprintf(esc_html__('The onOffice plugin has detected an active SEO plugin: %s. You currently have configured the onOffice plugin to fill out the title and description of the detail page, which can lead to conflicts with the SEO plugin.
+								We recommend that you go to the %s and configure the onOffice plugin to not modify the title and description. This allows you to manage the title and description with your active SEO plugin.', 'onoffice-for-wp-websites'), $listNamePluginSEO, $pluginOnofficeSetting);
+				$message = Parsedown::instance()
```
jmaas-onoffice, line 449: Please keep in mind injection vulnerabilities and use [Parsedown's safe mode](https://github.com/erusev/parsedown#security).

plugin/Gui/AdminPageApiSettings.php
```diff
@@ -175,31 +176,46 @@ private function addFormModelGoogleMapsKey()
...
+		}
+		if (count(array_intersect(["wordpress-seo/wp-seo.php"], get_option("active_plugins"))) > 0){
+			$pluginSEO['yoast_seo'] = 'Yoast SEO';
+		}
+		$listNamePluginSEO = implode(", ",$pluginSEO);
+		$titleDoNotModify = esc_html__("This plugin will not modify the title and description. This enables other plugins to manage those tags.",'onoffice-for-wp-websites');
+		$summaryDetailDoNotModify =  esc_html__('Available custom fields','onoffice-for-wp-websites');
+		$descriptionDetailDoNotModify = esc_html__('When this option is active, the plugin makes the following custom fields available in the detail view. These custom fields can be used in SEO plugins to fill out the title and description with the information of the currently shown estate. For information on how to use custom fields consult you SEO plugin\'s documentation.
+							These custom fields are only available in the detail view and on no other page.
+								\- Title (onoffice_title)
```
jmaas-onoffice, line 194: I would like the list to appear as an actual HTML list. So the proper Markdown syntax would be straight `-` without the escape `\-`.

grafik

I tried this and noticed that there are some CSS rules that remove the list styling. The Firefox inspector says the rules are coming from https://testinstanz-jmaa.onofficeweb.com/wp-admin/load-styles.php?c=1&dir=ltr&load%5Bchunk_0%5D=dashicons,admin-bar,common,forms,admin-menu,dashboard,list-tables,edit,revisions,media,themes,about,nav-menus,wp-pointer,widgets&load%5Bchunk_1%5D=,site-icon,l10n,buttons,wp-auth-check,wp-jquery-ui-dialog&ver=5.9.3. The result is this:

grafik

Could you check if you can make the list show up as in the first image? But don't try to fight the existing CSS (avoid `!important`). If it takes more than an hour, tell me and we'll find a different solution.

plugin/Controller/AdminViewController.php
```diff
@@ -427,19 +428,29 @@ public function generalAdminNoticeSEO() {
...
+		$pluginSEO = [];
+		if (count(array_intersect(["wpseo/wp-seo.php"], get_option("active_plugins"))) > 0) {
+			$pluginSEO['wp_seo'] = 'wpSEO';
+		}
+		if (count(array_intersect(["seo-by-rank-math/rank-math.php"], get_option("active_plugins"))) > 0) {
+			$pluginSEO['rank_math_seo'] = 'Rank Math SEO';
+		}
+		if (count(array_intersect(["wordpress-seo/wp-seo.php"], get_option("active_plugins"))) > 0) {
+			$pluginSEO['yoast_seo'] = 'Yoast SEO';
+		}
```
jmaas-onoffice, line 440: Would it make sense to extract this code into a function, since it is also used in https://github.com/onOfficeGmbH/oo-wp-plugin/pull/118/files/2c2659d9cd2c438696fa21e968c49a6863948678..3bb7947349d4816aecc4add7a78a875f5106a143?file-filters%5B%5D=.css&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.php&show-viewed-files=true#diff-c829b20424fa6035b35f0a1df38bda120f735f3eec5c0a2b5aa4a1004f06b973R180-R188?

jmaas-onoffice (changes requested):
A few notes and an addition.

The addition (it was not in the requirements) is to add some more SEO plugins. I've [searched for SEO plugins](https://wordpress.org/plugins/search/seo/) and would like to add those I found with over 100,000 installs.

The updated list is:
- `wordpress-seo/wp-seo.php` -> "Yoast SEO"
- `seo-by-rank-math/rank-math.php` -> "Rank Math SEO"
- `wpseo/wpseo.php` -> "wpSEO"
- `all-in-one-seo-pack/all_in_one_seo_pack.php` -> "All in One SEO"
- `autodescription/autodescription.php` -> "SEO Framework"
- `wp-seopress/seopress.php` -> "SEOPress"
- `squirrly-seo/squirrly.php` -> "Squirrly SEO"

plugin/Controller/AdminViewController.php
tang-hien-egs: sorry, i still working PR !

plugin/Controller/AdminViewController.php
yeneastgate: > A few notes and an addition.
> 
> The addition (it was not in the requirements) is to add some more SEO plugins. I've [searched for SEO plugins](https://wordpress.org/plugins/search/seo/) and would like to add those I found with over 100,000 installs.
> 
> The updated list is:
> 
> * `wordpress-seo/wp-seo.php` -> "Yoast SEO"
> * `seo-by-rank-math/rank-math.php` -> "Rank Math SEO"
> * `wpseo/wpseo.php` -> "wpSEO"
> * `all-in-one-seo-pack/all_in_one_seo_pack.php` -> "All in One SEO"
> * `autodescription/autodescription.php` -> "SEO Framework"
> * `wp-seopress/seopress.php` -> "SEOPress"
> * `squirrly-seo/squirrly.php` -> "Squirrly SEO"

@jmaas-onoffice I tested with The updated Seo plugin list with "custom field(4.2)" feature
A. Only this plugin allows support:
1.wordpress-seo/wp-seo.php -> "Yoast SEO" (ex:%%_cf_onoffice_title%%)

2.seo-by-rank-math/rank-math.php -> "Rank Math SEO"(ex:%customfield(onoffice_description)%)

3.wp-seopress/seopress.php -> "SEOPress"(ex:%%_cf_onoffice_title%%)

4.all-in-one-seo-pack/all_in_one_seo_pack.php -> "All in One SEO"(ex:Custom Field-onoffice_description )

B.This plugin allows do not support
5.wpseo/wpseo.php -> "wpSEO" 
(The sale of the wpSEO plugin was discontinued due to a change of ownership.
wpSEO will soon be free
In the future, wpSEP will be available free of charge with the same range of functions.)

6.autodescription/autodescription.php -> "SEO Framework"
 https://theseoframework.com/changelog/4-0-0-multiplex/
(When using custom meta title or meta description filters, they may not be processed for the meta input fields.
Fixing this requires restructuring the custom title and description methods with flags.)

7.squirrly-seo/squirrly.php -> "Squirrly SEO"(https://howto.squirrly.co/kb/seo-automation/#title_automation)

Now, will we remove this plugin do not support custom field?

yeneastgate: @jmaas-onoffice Please give me your opinion on this matter, thanks

jmaas-onoffice: So for wpSEO, SEO Framework, and Squirrly SEO we are not sure whether they can use custom fields?

It's actually ok if they can't. The warning is also about preventing double `<title>` and `<meta name="description">`. So it's worth keeping the warning for those plugins so that people can deactivate the plugin's `<title>` and `<meta>` so that there is only one. If their SEO plugin does not have the functions that they need, they will have to change to another SEO plugin.

wpSEO not being sold anymore is also ok. Some people might still have it and it is just one line in our code that can't break anything, right?

But thank you for your tests and research, especially the links to the documentation! It was a good idea to check it, I didn't know that some of them don't appear to support custom fields!

LongTrong-exe (approved):

hungnc89:
- Tester retested
- Video : 

Uploading Meta title and description für detail view1.mp4…

yeneastgate:
@jmaas-onoffice pls review again, thanks

plugin/Controller/AdminViewController.php
```diff
@@ -430,32 +432,72 @@ public function generalAdminNoticeSEO() {
...
+		return $listPluginSEOActive;
+	}
+	
+	public function displayUsingEmptyDefaultEmailError()
+	{
+		if ( ! get_option( 'onoffice-settings-default-email', '' ) && $this->getRecordManagerReadForm()->getCountDefaultRecipientRecord() != 0 ) {
+			$class                   = 'notice notice-error';
+			$label                   = __( 'plugin settings', 'onoffice-for-wp-websites' );
+			$defaultEmailAddressLink = sprintf( '<a href="admin.php?page=onoffice-settings">%s</a>', $label );
+			/* translators: %s will be replaced with the translation of 'plugin settings'. */
```
jmaas-onoffice, line 491: Very nice! That's really useful information!

jmaas-onoffice (changes requested):

tests/TestClassAdminViewController.php
```diff
@@ -295,7 +295,7 @@ public function testGeneralAdminNoticeSEO()
 		$pAdminViewController = new AdminViewController();
 		$pAdminViewController->generalAdminNoticeSEO();
 		$this->expectOutputString("<div class=\"notice notice-warning active-plugin-seo is-dismissible\"><p>The onOffice plugin has detected an active SEO plugin: Yoast SEO. You currently have configured the onOffice plugin to fill out the title and description of the detail page, which can lead to conflicts with the SEO plugin.<br />
-We recommend that you go to the <a href='http://example.org/wp-admin/admin.php?page=onoffice-settings' target='_blank' rel='noopener'>onOffice plugin settings</a> and configure the onOffice plugin to not modify the title and description. This allows you to manage the title and description with your active SEO plugin.</p></div>");
+We recommend that you go to the <a href='<a href=\"http://example.org/wp-admin/admin.php?page=onoffice-settings\">http://example.org/wp-admin/admin.php?page=onoffice-settings</a>' target='_blank' rel='noopener'>onOffice plugin settings</a> and configure the onOffice plugin to not modify the title and description. This allows you to manage the title and description with your active SEO plugin.</p></div>");
```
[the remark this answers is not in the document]
yeneastgate: yes, i will check it again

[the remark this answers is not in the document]
yeneastgate: @jmaas-onoffice , I fixed it, pls review again, thanks

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
1. When I set the option to "Do not modify" and then open a detail page, the title is still set to estate's title. It should be left as the WordPress page's title when I do not have SEO plugins active.

    The point of having this option is to prevent conflicts where both our plugin and a SEO plugin are trying to set the title and description. So if our plugin is still setting the title when the option is set to "Do not modify", then we might run into conflicts. So please make sure that with this setting the plugin actually does not modify those tags.

2. Minor improvement: Could we add a jump mark to the setting, so that when you click on the link in the warning, it scrolls to the title and description setting? One simple way to achieve this would be to give a HTML ID to the setting and then add a fragment to the link (like `#title-and-description`).

yeneastgate:
> 1. When I set the option to "Do not modify" and then open a detail page, the title is still set to estate's title. It should be left as the WordPress page's title when I do not have SEO plugins active.
>    The point of having this option is to prevent conflicts where both our plugin and a SEO plugin are trying to set the title and description. So if our plugin is still setting the title when the option is set to "Do not modify", then we might run into conflicts. So please make sure that with this setting the plugin actually does not modify those tags.
> 2. Minor improvement: Could we add a jump mark to the setting, so that when you click on the link in the warning, it scrolls to the title and description setting? One simple way to achieve this would be to give a HTML ID to the setting and then add a fragment to the link (like `#title-and-description`).

yes, I will check it.

jmaas-onoffice:
To clarify the first point:

I didn't realize that with v3.2 the plugin already modifies the title. I think in a previous version, all detail pages had the same title, namely the title of the WordPress page, ususally something like "Estate details". But it seems that that is no longer the case, sorry for the confusion!

As I explained, the goal is to prevent conflicts where our plugin and a SEO plugin both try to modify the `<title>` and other tags. But the only way each detail page can have a different title corresponding to the estate's title from onOffice enterprise is if the plugin is modifying the `<title>`.

So could you have a look at whether our plugin is modifying the `<title>` even if the option is set to "Do not modify"?

yeneastgate:
> To clarify the first point:
> 
> I didn't realize that with v3.2 the plugin already modifies the title. I think in a previous version, all detail pages had the same title, namely the title of the WordPress page, ususally something like "Estate details". But it seems that that is no longer the case, sorry for the confusion!
> 
> As I explained, the goal is to prevent conflicts where our plugin and a SEO plugin both try to modify the `<title>` and other tags. But the only way each detail page can have a different title corresponding to the estate's title from onOffice enterprise is if the plugin is modifying the `<title>`.
> 
> So could you have a look at whether our plugin is modifying the `<title>` even if the option is set to "Do not modify"?

yes, i will check it again

hungnc89:
- Tester test
- Vide

https://user-images.githubusercontent.com/72601287/173037714-1d27a90e-9734-4194-98dc-a7daa91f1781.mp4

o

yeneastgate:
> > To clarify the first point:
> > I didn't realize that with v3.2 the plugin already modifies the title. I think in a previous version, all detail pages had the same title, namely the title of the WordPress page, ususally something like "Estate details". But it seems that that is no longer the case, sorry for the confusion!
> > As I explained, the goal is to prevent conflicts where our plugin and a SEO plugin both try to modify the `<title>` and other tags. But the only way each detail page can have a different title corresponding to the estate's title from onOffice enterprise is if the plugin is modifying the `<title>`.
> > So could you have a look at whether our plugin is modifying the `<title>` even if the option is set to "Do not modify"?
> 
> yes, i will check it again

I checked again, You can see video https://user-images.githubusercontent.com/72601287/173037714-1d27a90e-9734-4194-98dc-a7daa91f1781.mp4
our plugin **do not modify** the `<title>` even if the option is set to "Do not modify".
Only when we add customfield to seo plugin then Title will modify.
I fixed link scrolls to the title and description setting.

jmaas-onoffice (approved):

jmaas-onoffice:
I have trouble reproducing the problem with the title. Sometimes it still appears, but sometimes when I do the same thing it does not. So I'll need some time to figure out what's going on.

jmaas-onoffice:
I'm incredibly confused. I have a video, where I managed to have the title behave as desired. But when I try to recreate that now, it never works. I don't really know what to do...

So in this video with v2.20.0, notice how the title is the same for all estates and it is the one set for the detail page:

https://user-images.githubusercontent.com/90763364/173596119-87a9abd2-470d-453e-a8c8-9aaa2d94dfd2.mp4

It seems that since the title-in-URL feature was added in v2.20.1, the title of the page might have already been set to the title of the estate. But only the plugin knows this information, so this means that the plugin must change the title. It should be like in the video, where we have the WordPress title.

Does this make it a bit clearer, what I have in mind?

yeneastgate:
> I'm incredibly confused. I have a video, where I managed to have the title behave as desired. But when I try to recreate that now, it never works. I don't really know what to do...
> 
> So in this video with v2.20.0, notice how the title is the same for all estates and it is the one set for the detail page:
> 
>  wordpress-page-title.mp4 
> It seems that since the title-in-URL feature was added in v2.20.1, the title of the page might have already been set to the title of the estate. But only the plugin knows this information, so this means that the plugin must change the title. It should be like in the video, where we have the WordPress title.
> 
> Does this make it a bit clearer, what I have in mind?

yes, I will test it again

hungnc89:
- I tested it on your server and it is working
- Video:

https://user-images.githubusercontent.com/72601287/173725947-185c332f-e522-40bc-a39f-85da55465eee.mp4

yeneastgate:
@jmaas-onoffice I found the cause.
Pls check lines 135->137 in "plugin.php" file
image
Jakob has made a function to change the title.
do we keep parallelism or remove it?
You just need to remove it, the issue when you select the option "Do not modify" but do not active plugin SEO will disappear. The title will do not modify.

jmaas-onoffice:
@yeneastgate Very nice, thanks for investigating this!

You can decide whether to remove this. The goal is that when the user selects "Do not modify" the title stays the same and when they select "Fill out" that the plugin sets the title and meta description. :)

yeneastgate:
> @yeneastgate Very nice, thanks for investigating this!
> 
> You can decide whether to remove this. The goal is that when the user selects "Do not modify" the title stays the same and when they select "Fill out" that the plugin sets the title and meta description. :)

yes I will check it again and will decide whether to remove this

plugin/Controller/EstateViewDocumentTitleBuilder.php
```diff
@@ -43,29 +43,20 @@ public function __construct(EstateTitleBuilder $pEstateTitleBuilder, WPQueryWrap
...
-	public function buildDocumentTitle(array $title): array
+	public function buildDocumentTitleField(string $format): string
 	{
 		$estateId = (int)$this->_pWPQueryWrapper->getWPQuery()->get('estate_id', 0);
 		if ($estateId === 0) {
-			return $title;
+			return '';
 		}
-
-		$title['title'] = $this->_pEstateTitleBuilder->buildTitle($estateId, '%1$s');
```
jmaas-onoffice, line 58: With the current code, the plugin replaces the entrie `<title>` with just the estate's title, e.g. "Beautiful estate". It would be better if the title would still contain the title separator and site name like "Beautiful estate | My Real Estate".

I think one way of achieving this is using the [`document_title_parts` filter](https://developer.wordpress.org/reference/hooks/document_title_parts/), like this previous code did, and just replacing the `$title['title']` so that the other parts are kept.

jmaas-onoffice (changes requested):
I think this looks pretty good now! We're almost there!

I noticed that I can't use the custom field for `onoffice_id`/`onoffice_datensatznr`. When I try to add it via Yoast or Rank Math, it is empty. So `%%cf_onoffice_id%% %%cf_onoffice_title%%` in Yoast displays as "Very nice estate" instead of "431 Very nice estate". Could you check that all custom fields listed in the settings can be actually used?

plugin/Controller/EstateViewDocumentTitleBuilder.php
yeneastgate: yes, i will check and fix it.

yeneastgate: @jmaas-onoffice  I checked the custom field for onoffice_id/onoffice_datensatznr work fine.Note You need add
Screenshot
:

yeneastgate: > With the current code, the plugin replaces the entrie `<title>` with just the estate's title, e.g. "Beautiful estate". It would be better if the title would still contain the title separator and site name like "Beautiful estate | My Real Estate".
> 
> I think one way of achieving this is using the [`document_title_parts` filter](https://developer.wordpress.org/reference/hooks/document_title_parts/), like this previous code did, and just replacing the `$title['title']` so that the other parts are kept.

I think it is not necessary because All SEO plugins have variable for separator ex: Yoast "%%sep%%"

LongTrong-exe: @jmaas-onoffice What title format do you want to use?
"Title | Site Name" or like the old one with separator

image

jmaas-onoffice: I think it's best to use [document_title_parts filter](https://developer.wordpress.org/reference/hooks/document_title_parts/) as described above, because then you only need to set the estate title and the rest is handled by WordPress.

jmaas-onoffice:
Sorry, I was too slow to review this. Could you update this branch and fix the merge conflicts?

dai-eastgate:
Updated. Please check it.

jmaas-onoffice (approved):
I will test it.

jmaas-onoffice (changes requested):
1. By default, I have "-" as the title separator.
    grafik

    When I set the plugin to "Fill Out", then the title separator changes:
    
    grafik

    Is this intentional? What is the motivation?

2. I forgot an important field in the requirements! In addition to the internal database number, the user should be able to use the externally visible property number ("Prop No.", `objektnr_extern`). Could you add that as an available custom field as described in the updated requirements (#223)?

Looks good otherwise. :)

LongTrong-exe:
@jmaas-onoffice I set the separator to `|` as you described in this comment https://github.com/onOfficeGmbH/oo-wp-plugin/pull/118#discussion_r900140618 :sweat_smile:

jmaas-onoffice:
Oh, I didn't mean that we should set the separator, sorry! I just meant that we should keep the existing one. :)

Could you change it so that it only modifies the title part?

LongTrong-exe (approved):

jmaas-onoffice:
It looks good so far. I just realized that we might do a change to the custom fields before we release this to prevent a breaking change. I still have to think about this, though...

LongTrong-exe (dismissed):

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
1. The description in the settings is missing the \<details>.
    grafik
2. The warning with conflicting SEO plugins does not appear anymore in the settings:
    grafik
    It does appear on the other pages, but the settings page is broken.
3. Yoast seems to have trouble with dashes in custom fields. It does not change the appearance of the custom field and the user might think it did not work:
    grafik
    The custom field actually works this way, but we can change the syntax so that Yoast recognizes it. I have updated 4.3 in #223.
4. Change of requirements: It would be better to shorten strings at whitespace instead of in the middle of the word. This is also updated in 4.3 in #223.

LongTrong-exe (dismissed):

plugin.php
```diff
@@ -167,21 +167,26 @@
 }
 
 // Return title custom by custom field onOffice
-function getRestrictLength( $valueNumber, $title ) {
-	if ( empty( $valueNumber ) ) {
+function getRestrictLength( $limitEllipsis, $title ,$ellipsis = '...' ) {
```
jmaas-onoffice, line 170: Please use the proper ellipsis character, `…`, instead of three periods.

jmaas-onoffice (changes requested):
Some issues from https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/118#pullrequestreview-1062973152 are not fixed.

1. is still the same. For 2. it's actually that the styling of the warning is missing (that's why I thought it was missing, but the message is actually there in the screenshot).

plugin.php
dai-eastgate: For 2. you can manually clear the cache which will fix that.

jmaas-onoffice (approved):
For creating a dev build.

jmaas-onoffice (changes requested):
The previous issues are still present:
grafik

Is it clear what the problem is?

jmaas-onoffice:
Oh, so the warning was a caching issue? Next time, you can also tell me that I should reset my cache, I forgot. ;)

So there is only the issue custom fields description left. Is it clear what I mean by that? :)

jmaas-onoffice (changes requested):
There still is a `<summary>` element without a `<detail>` element.

https://github.com/onOffice-Web-Org/oo-wp-plugin/blob/b974d562b23aea8cd6d99f55fe097787e80eb0b3/plugin/Gui/AdminPageApiSettings.php#L235

jmaas-onoffice (approved):

jmaas-onoffice (dismissed):

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
I think we're really close now!

Some of the enterprise fields have underscores in their keys, e.g. `ausstatt_beschr` or `sonstige_angaben`. It seems that I can't output those in Yoast with the ellipsis feature.

So when I modify Yoast's description of the detail page and I add `cf_onoffice_sonstige_angaben`, then it outputs the whole thing. When I instead use `cf_onoffice_ellipsis50_sonstige_angaben`, nothing gets output.

Could you have a look and fix that?

dai-eastgate:
@jmaas-onoffice I fixed it. Please review again

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
It all works great!

But I thought of an issue with upgrading: If someone has an SEO plugin active and they update the plugin through automation, they might not see the warning.

Therefore it would be better to set the default setting to "Fill out" only when no SEO plugin is detected, and to "Do not modify" when we detect such a plugin.

Could you add this to the upgrade logic?

dai-eastgate:
https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/118#pullrequestreview-1131254541
I added it you can review. :)

dai-eastgate:
@jmaas-onoffice I added it you can review. :)

jmaas-onoffice (approved):
I'll have to test it to see if it works as intended.

jmaas-onoffice:
This currently seems to always set the setting to "Fill out". Even when on the old version I activate Yoast and then upgrade, it still sets "Fill out".

Maybe the requirement in https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/118#pullrequestreview-1131254541 is not clear. Do you have any questions?

jmaas-onoffice (changes requested):
This currently seems to always set the setting to "Fill out". Even when on the old version I activate Yoast and then upgrade, it still sets "Fill out".

Maybe the requirement in https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/118#pullrequestreview-1131254541 is not clear. Do you have any questions?

dai-eastgate:
Yes i understood and fixed it

jmaas-onoffice:
Ok, this solution works for a fresh install, but it should also work when people upgrade.

Imagine someone has Yoast active and they automatically upgrade. Currently the plugin actually still sets it to "Fill out", so it might overwrite their Yoast settings. So the logic of checking whether SEO plugins are active should also apply when upgrading.

jmaas-onoffice (changes requested):
Ok, this solution works for a fresh install, but it should also work when people upgrade.

Imagine someone has Yoast active and they automatically upgrade. Currently the plugin actually still sets it to "Fill out", so it might overwrite their Yoast settings. So the logic of checking whether SEO plugins are active should also apply when upgrading.

dai-eastgate:
@jmaas-onoffice Sorry, I missed your comment. I have just fixed, could you please review.

jmaas-onoffice:
@dai-eastgate No worries! :)

plugin/Controller/AdminViewController.php
```diff
@@ -443,6 +443,9 @@ public function generalAdminNoticeSEO() {
 		$listPluginSEOActive = $this->getPluginSEOActive();
 		$listNamePluginSEO = implode(", ", $listPluginSEOActive);
 		if (count($listPluginSEOActive) > 0) {
+			if(get_option('onoffice-settings-title-and-description') === null){
```
jmaas-onoffice, line 446: Is this code still necessary?

jmaas-onoffice (changes requested):

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
Now when I upgrade without Yoast active, it still sets "Do not modify" but it should set it to "Fill out".

dai-eastgate:
@jmaas-onoffice i fixed it, could you review?

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
It feels like we're really close, but I have found another bug:

When the detail page includes a contact form and the contact form has "Set Estate Context as Heading" active, then the detail page crashes:

```
Fatal error: Uncaught Error: Return value of onOffice\WPlugin\Controller\EstateTitleBuilder::buildEstateTitle() must be of the type string, bool returned
in /www/htdocs/w01bb2af/testinstanz-jmaa.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Controller/EstateTitleBuilder.php on line 161
```

dai-eastgate:
@jmaas-onoffice I fixed it, can you check it again?

PR #121, opened 2021-10-12
P#54965 "vermarktungsstatus" is added to detail view during installation and leads to error

vu-vuong-esg:
P#54965 "vermarktungsstatus" is added to detail view during installation and leads to error.


plugin/ViewFieldModifier/ViewFieldModifierHandler.php
```diff
@@ -108,6 +108,9 @@ public function getAllAPIFields(): array
 		}
 
 		$fields = array_values(array_unique($apiFields));
+		if (($key = array_search('vermarktungsstatus', $fields)) !== false) {
```
jayay, line 111: I am confused. If it appears here, that would be a logic error. Can this actually happen?

PR #122, opened 2021-10-12
P#55021 Use onOffice logo in menu

vu-vuong-esg:
P#55021 Use onOffice logo in menu


plugin/Controller/AdminViewController.php
```diff
@@ -148,7 +148,7 @@ public function register_menu()
 		add_menu_page( __('onOffice', 'onoffice-for-wp-websites'), __('onOffice', 'onoffice-for-wp-websites'),
 			$roleMainPage, $this->_pageSlug, function() use ($pDI) {
 				echo $pDI->get(MainPage::class)->render();
-			}, 'dashicons-admin-home');
+			}, plugin_dir_url(ONOFFICE_PLUGIN_DIR.'/index.php').'images/onOffice-Logo-screen_wp-grey.svg');
```
jayay, line 151: If this works, the behavior is not documented: https://developer.wordpress.org/reference/functions/add_menu_page/#parameters

It would be safer to use the base64 variant.

css/admin.css
```diff
@@ -259,4 +259,11 @@ div[data-name^=oopluginfieldconfigformdefaultsvalues-value].multiselect > input.
 	width: 50%;
 	text-align: left;
 	margin-top: 0;
+}
+#adminmenu .wp-menu-image img {
```
jayay, line 263: This affects all the images, right?

PR #132, opened 2021-11-09
P#57353 [WP-Plugin] Human-readable URLs showing ID and title

hungnc89:


michaelarnold-dev:
@tung-le-esg Could you please check this as well?

jayay:
Fixes for the broken tests were [committed to master](https://github.com/onOfficeGmbH/oo-wp-plugin/compare/5ef5d1fa0b8f344b5332d551cdbe996871866bdb..2a4ff1fdc14c863308695fefc2f6affe641c3b09) today. These simply need to merged into this branch.

jayay (approved):

tests/TestClassRedirectIfOldUrl.php
```diff
@@ -75,6 +75,7 @@ public function testGetUri()
 	 */
 	public function testRedirectDetailViewSameUrl()
 	{
+		add_option('onoffice-detail-view-showTitleUrl', true);
```
jayay, line 78: Both with and without that option should be tested

hoangdoegs: Hi @jayay,
In this case I cannot assert for `wp_redirect`. I think WordPress does not support this case yet

jayay: Yeah, a wrapper class for `wp_redirect` would be needed, which in turn could be mocked for the test.

plugin/Utility/RedirectIfOldUrl.php
```diff
@@ -5,34 +5,38 @@
...
+	private $_wpPageWrapper;
+	private $_wpRedirectWrapper;
 
 	public function __construct()
 	{
 		$pContainerBuilder = new ContainerBuilder;
 		$pContainerBuilder->addDefinitions(ONOFFICE_DI_CONFIG_PATH);
 		$pContainer = $pContainerBuilder->build();
 		$this->_pLanguageSwitcher = $pContainer->get(EstateDetailUrl::class);
+		$this->_wpPageWrapper = $pContainer->get(WPPageWrapper::class);
```
jayay, line 23: It'd be cleaner and easier to test if the dependencies were obtained as constructor parameters:

```php
	public function __construct(EstateDetailUrl $pEstateDetailUrl, WPPageWrapper $pWPPageWrapper, WPRedirectWrapper $pWPRedirectWrapper)
	{
		$this->_pLanguageSwitcher = $pEstateDetailUrl;
		$this->_wpPageWrapper = $pWPPageWrapper;
		$this->_wpRedirectWrapper = $pWPRedirectWrapper;
	}
```

plugin/Utility/RedirectIfOldUrl.php
```diff
@@ -50,4 +54,9 @@ public function getCurrentLink(): string
 		global $wp;
 		return home_url(add_query_arg(array(), $wp->request));
 	}
+
+	public function setRedirectWrapper($redirectWrapper)
```
jayay: this setter needs to go. The constructor change would solve this problem.

tests/TestClassRedirectIfOldUrl.php
```diff
@@ -92,6 +94,32 @@ public function testRedirectDetailViewSameUrl()
...
+			'post_author' => 1,
+			'post_content' => '[oo_estate view="detail"]',
+			'post_title' => 'Detail View',
+			'post_type' => 'page',
+		]);
+		$redirect = $this->getMockBuilder(WPRedirectWrapper::class)
+			->setMethods(['redirect'])
+			->getMock();
+		$redirect->method('redirect')->willReturn(true);
+		$this->_pRedirectIfOldUrl->setRedirectWrapper($redirect);
```
jayay: please remove this line

jayay (commented):
Apart from my other comments, could you please also add doc blocks above methods that have arguments and return types and also above member variables? Thank you

plugin/Utility/RedirectIfOldUrl.php
hoangdoegs: hi @jayay,
I need setter functions to set mock for the test

jayay: Just use the constructor! That'd be a lot cleaner.

plugin/EstateList.php
```diff
@@ -119,6 +121,34 @@ public function __construct(DataView $pDataView, EstateListEnvironment $pEnviron
 		$this->_pLanguageSwitcher = $pContainer->get(EstateDetailUrl::class);
 	}
 
+	public function redirectUrlValid()
```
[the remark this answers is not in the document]
jayay: the name `redirectIfOldUrl` was meanto to be for the method, not the class... 😬

plugin/Utility/RedirectIfOldUrl.php
hoangdoegs: @jayay constructor accepts params on the correct instance. eg WPPageWrapper. But when I want to pass the mock of WPPageWrapper, eg WPPageWrapperMocked, the constructor gives an error

hoangdoegs: @jayay constructor accepts params on the correct instance. eg WPPageWrapper. But when I want to pass the mock of WPPageWrapper, eg WPPageWrapperMocked, the constructor gives an error

jayay: It should work, as long as the new mock is a subclass of `WPPageWrapper`. Mocks in PHPUnit inherit from the mocked class. What exactly is the error?

tests/TestClassRedirectIfOldUrl.php
```diff
@@ -97,39 +101,27 @@ public function testRedirectDetailViewSameUrl()
...
-
-		 $this->assertTrue($redirectIfOldUrl->redirectDetailView(1, 1, 'tes post'));
+		$pWPPost = self::factory()->post->create_and_get([
+			'post_author' => 1,
+			'post_content' => '[oo_estate view="detail"]',
+			'post_title' => 'Detail View',
+			'post_type' => 'page',
+		]);
+		$pLanguageSwitcher = new EstateDetailUrl();
+		$wpRedirectWrapperMocker = new RedirectWrapperMocker();
```
jayay, line 117: Mock could be built using PHPUnit, instead of making a new class (which should also be tested)

plugin/Utility/Redirector.php
```diff
@@ -7,21 +7,18 @@
...
-	public function __construct()
+	public function __construct(EstateDetailUrl $estateDetailUrl, WPPageWrapper $pageWrapper, WPRedirectWrapper $redirectWrapper)
 	{
-		$pContainerBuilder = new ContainerBuilder;
-		$pContainerBuilder->addDefinitions(ONOFFICE_DI_CONFIG_PATH);
-		$pContainer = $pContainerBuilder->build();
-		$this->_pLanguageSwitcher = $pContainer->get(EstateDetailUrl::class);
-		$this->_wpPageWrapper = $pContainer->get(WPPageWrapper::class);
-		$this->_wpRedirectWrapper = $pContainer->get(WPRedirectWrapper::class);
+		$this->_pLanguageSwitcher = $estateDetailUrl;
```
jayay, line 19: Please also rename this variable

jayay (commented):
Almost done! Just two small things

tests/TestClassRedirectIfOldUrl.php
hoangdoegs: @jayay Please review the [above](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/132#discussion_r776669095). Because you asked me to write Mocker Class

jayay: Either way works

jayay (approved):

michaelarnold-dev:
@jayay Can that now be tested by our QA-team or is there still something to to?

jayay:
I'm giving this to QA

PR #144, opened 2021-11-23
P#50179 Show or hide reference estate in estate list

hungnc89:
Update query add column in estate list


jayay:
@hungnc89 I think merge-commit 9120d9d went wrong here, at least it is not recognized as a merge commit. Fixing this could probably only done by creating a new branch and then cherry-picking the necessary commits into that new branch. Would you please consider trying that out?

PR #146, opened 2021-11-26
P#59183 Message can be selected twice in the contact form 18668

hungnc89:
filter and remove (unnamed category) contains message field


hungnc89:
@jayay can you review this PR?

jayay:
Message gets added to the field list once in https://github.com/onOfficeGmbH/oo-wp-plugin/blob/master/plugin/Field/FieldModuleCollectionDecoratorFormContact.php but where's the other class that adds it? I think it'd be better to fix the problem at the root and find out where else the message gets added to the field list.

hungnc89:
@jayay Can you review my commit and give some suggestions? I am a bit dissatisfied with this solution but I have no other solution.

plugin/Gui/AdminPageBase.php
```diff
@@ -109,6 +110,47 @@ protected function addFormModel(FormModel $pFormModel)
...
 
+	/**
+	 * @param $slug
+	 * @param string $removeField
+	 * @return void
+	 */
+
+	protected function removeFieldsInFormModel($slug, string $removeField)
+	{
+		$inputModels = $this->_formModels[$slug]->getInputModel();
```
jayay, line 121: it is generally prettier if code doesn't act on class members or global state

jayay (commented):
I think the main problem is, that the message field comes from here:
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/192db62085f0795cae609bca95762e2cacfefc4a/plugin/Model/FormModelBuilder/FormModelBuilderDBForm.php#L92
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/192db62085f0795cae609bca95762e2cacfefc4a/plugin/Model/FormModelBuilder/FormModelBuilderDBForm.php#L149
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/192db62085f0795cae609bca95762e2cacfefc4a/plugin/Field/Collection/FieldsCollectionBuilderShort.php#L108
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/192db62085f0795cae609bca95762e2cacfefc4a/plugin/Field/FieldModuleCollectionDecoratorFormContact.php#L49

and here:
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/192db62085f0795cae609bca95762e2cacfefc4a/plugin/Gui/AdminPageFormSettingsBase.php#L458

But it also gets added here:
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/192db62085f0795cae609bca95762e2cacfefc4a/plugin/Gui/AdminPageFormSettingsContact.php#L179

This is too complicated and needs to be simplified. Also, `FieldsCollection` does not seem to check the `$_fields` array for uniqueness (maybe use the field name as the key there and do an `array_values()`  in the getter?)

hungnc89:
- Tester tested
- Video:

https://user-images.githubusercontent.com/72601287/163985210-a08edd58-ab82-4b89-a1b2-777644fc42ba.mp4

jmaas-onoffice (approved):

PR #154, opened 2021-12-17
P#60599 Space for map is taken even when the map is not shown

tung-le-esg:
P#60599 Space for map is taken even when the map is not shown


hungnc89:

jayay (approved):

PR #158, opened 2021-12-29
P#50179 Show or hide reference estate in estate lis

hoangdoegs:
Recreate pull request


jayay (approved):

michaelarnold-dev:
@nglelinh Test in https://wpptest.onofficeweb.com/ with version 2.20.12-1-gff712798:

When trying to save a property list, I always get the error message: "An error occurred while saving the view. Please make sure that no other view with this name exists, even if it has a different type."

->> Property lists should be able to be saved. There is no view with the same name here.

hoangdoegs:
@jayay I checked it.
Because db_version hasn't been updated yet, the query add column has not been run.
Please check again

jayay (approved):

michaelarnold-dev:
@nglelinh Test with version 2.21.6-4-gc171e430, https://wpptest.onofficeweb.com/

The errors in #2754903 and #2754937 are not fixed for me.

1.) Requirement: By default the checkbox should be selected. This way the change remains downward compatible.
--> In existing and new property lists the checkbox is not selected.

2.) Requirement: If the new checkbox in backend is not selected: The reference properties are filtered out from the list view and no longer be displayed.
->> Removing the checkbox has no effect on me. I.e. reference properties are displayed in the frondend although the checkbox is unchecked.

jayay (approved):

PR #104, opened 2021-08-12
P#52017  fix php di issue and type errors

vu-vuong-esg:

PR #159, opened 2021-12-29
P#51853 Fix the incompatibility between WP 5.8 and PHP 8

hoangdoegs:


jayay:
P#52017 / #104 deals with this. It wasn't published yet, because that would be a major release.

jmaas-onoffice:
Since this PR is quite outdated, it seems easier to create a new one. I've opened #219.

PR #160, opened 2021-12-31
P#61223 19600 reload edit pages after saving and show copyable shortcode there

vu-vuong-esg:


hungnc89:

hungnc89:
@jayay please check PR help me, thanks

hungnc89:

js/onoffice-copycode.js
```diff
@@ -0,0 +1,9 @@
+jQuery(document).ready(function($){
+    $('#button-copy').click(function() {
+        var copyCode = document.getElementsByTagName("code").item(0);
+        // console.log(copyCode);
```
jayay, line 4: please remove the debug code

plugin/Model/FormModelBuilder/FormModelBuilderDBEstateListSettings.php
```diff
@@ -123,7 +124,32 @@ public function generate(string $pageSlug, $listViewId = null): FormModel
...
+	public function createInputModelEmbedCode()
+	{
+		$pConfig = new InputModelDBFactoryConfigEstate();
+		$config = $pConfig->getConfig();
+		$name = $config[InputModelDBFactory::INPUT_LISTNAME]
+		[InputModelDBFactoryConfigEstate::KEY_FIELD];
+
+		$listName = $this->getValue($name);
 
+		$code = '[oo_estate view=="'.$listName.'"]';
```
jayay, line 139: one `=` too much

plugin/Renderer/InputFieldButtonRenderer.php
```diff
@@ -0,0 +1,79 @@
...
+
+	/**
+	 *
+	 */
+
+	public function render()
+	{
+		$id = HtmlIdGenerator::generateByString($this->_id);
+		echo '<input type="button" class="button" id="button-copy" name="'
+			.esc_attr($id).'" value="'.esc_html__('Copy !', 'onoffice-for-wp-websites').'" '
```
jayay, line 58: `Copy!` would be better

jayay (changes requested):

vu-vuong-esg:
@jayay i fixed, can you check PR help me ?

plugin/Model/FormModelBuilder/FormModelBuilderDBEstateListSettings.php
```diff
@@ -123,7 +124,32 @@ public function generate(string $pageSlug, $listViewId = null): FormModel
...
+	public function createInputModelEmbedCode()
+	{
+		$pConfig = new InputModelDBFactoryConfigEstate();
+		$config = $pConfig->getConfig();
+		$name = $config[InputModelDBFactory::INPUT_LISTNAME]
+		[InputModelDBFactoryConfigEstate::KEY_FIELD];
+
+		$listName = $this->getValue($name);
 
+		$codes = '[oo_estate view=="'.$listName.'"]';
```
jayay, line 139: one `=` too much here

tests/TestClassFormModelBuilderDBEstateListSettings.php
```diff
@@ -215,4 +216,29 @@ public function testCreateInputModelShowReferenceEstate()
...
+		$pInstance->method('getInputModelDBFactory')->willReturn($this->_pInputModelFactoryDBEntry);
+		$pInstance->method('getValue')->willReturn('1');
+		$pInputModelFormEmbedCode = $pInstance->createInputModelEmbedCode();
+		$this->assertInstanceOf(InputModelLabel::class, $pInputModelFormEmbedCode);
+		$this->assertEquals($pInputModelFormEmbedCode->getHtmlType(), 'label');
+	}
+
+	public function testCreateInputModelButton()
+	{
+		$pInstance = $this->getMockBuilder(FormModelBuilderDBEstateListSettings::class)
```
jayay, line 236: Testing a class using a mock of itself doesn't really make sense

jayay (changes requested):

tests/TestClassFormModelBuilderDBEstateListSettings.php
tang-hien-egs: i fixed it, please check PR help me !

tests/TestClassFormModelBuilderDBForm.php
```diff
@@ -54,22 +57,26 @@ public function prepare()
 		$pContainerBuilder = new ContainerBuilder;
 		$pContainerBuilder->addDefinitions(ONOFFICE_DI_CONFIG_PATH);
 		$this->_pContainer = $pContainerBuilder->build();
+		$this->_pInstance = $this->getMockBuilder(FormModelBuilderDBForm::class)
```
jayay, line 60: The test subject is a mock again. Only the test subjects' _dependencies_ are supposed to be a mocks, if needed (that means, if they have a side-effect that needs to be avoided). Please replace the MockBuilder with `new FormModelBuilderDBForm`.

tests/TestClassFormModelBuilderDBForm.php
```diff
@@ -152,4 +159,54 @@ public function testGetDataContactTypeThrowException()
 
 		$this->assertEquals([], $result);
 	}
+	public function testCreateInputModelName()
```
jayay, line 162: Thank you for adding more cases!

jayay (changes requested):

jayay (dismissed):

jayay (approved):

michaelarnold-dev:
@LongTrong-exe 
Testing  feedback by Robert Igelmund:

1.) According to the requirements, the shortcode should be shown for lists and forms in the detail views. This is not the case for the address list and the unit list.
2.) The title for forms should be as in list_desired.png. There you can see "Type: Interest Form...". "Type of form: Interest Form..." was implemented.
list_desired

vu-vuong-esg:
> 

@michaelarnold-dev , i will fix it as soon as

tang-hien-egs:
@michaelarnold-dev , i fixed it, could you check PR help me ?

jayay (approved):

michaelarnold-dev:
@tang-hien-egs We identified a new error when calling the address detail view:

Fatal error: Uncaught TypeError: Argument 1 passed to onOffice\WPlugin\Gui\AdminPageAjax::createMetaBoxByForm() must be an instance of onOffice\WPlugin\Model\FormModel, null given, called in /www/htdocs/w01b9fbd/wpptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Gui/AdminPageAddressListSettings.php on line 196 and defined in /www/htdocs/w01b9fbd/wpptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Gui/AdminPageAjax.php:89 Stack trace: #0 /www/htdocs/w01b9fbd/wpptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Gui/AdminPageAddressListSettings.php(196): onOffice\WPlugin\Gui\AdminPageAjax->createMetaBoxByForm(NULL, 'side') #1 /www/htdocs/w01b9fbd/wpptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Gui/AdminPageSettingsBase.php(171): onOffice\WPlugin\Gui\AdminPageAddressListSettings->generateAccordionBoxes() #2 /www/htdocs/w01b9fbd/wpptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/pl in /www/htdocs/w01b9fbd/wpptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Gui/AdminPageAjax.php on line 89

Could you please analyze?

tang-hien-egs:
> 

@michaelarnold-dev , this is error exist in PR master, i fixed it in #172

michaelarnold-dev:
@tang-hien-egs According to the requirements, the shortcode should be shown for lists and forms in the detail views. This is not the case for the address list and the unit list."

tang-hien-egs:
> @tang-hien-egs According to the requirements, the shortcode should be shown for lists and forms in the detail views. This is not the case for the address list and the unit list."

video test : shortcode not display in the address list and unit list:

https://user-images.githubusercontent.com/95266110/164360686-df76212b-f036-46ec-a085-6c946c39f287.mp4

tang-hien-egs:
@michaelarnold-dev , could you check PR help me ?

michaelarnold-dev:
@tang-hien-egs I assigned it to Robert Igelmund. He has to check it

tang-hien-egs:
> @tang-hien-egs I assigned it to Robert Igelmund. He has to check it

yes, thank you !

michaelarnold-dev:
@tang-hien-egs According to Robert the short codes should be visable for address and unit list. He writes "According to the requirements, the shortcode should be shown for lists and forms in the detail views. This is not the case for the address list and the unit list"

tang-hien-egs:
> @tang-hien-egs According to Robert the short codes should be visable for address and unit list. He writes "According to the requirements, the shortcode should be shown for lists and forms in the detail views. This is not the case for the address list and the unit list"

@michaelarnold-dev , i checked with latest source code, shortcode not show in address and unit list. i will discuss with Robert to clean issue !

vu-vuong-esg:
@jmaas-onoffice  i fixed conflict, please check PR help me !

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
The notice "The form was saved" (or the equivalent for lists) should still be displayed. With the current code it is displayed very quickly before the reload and is then gone.

tang-hien-egs:
> The notice "The form was saved" (or the equivalent for lists) should still be displayed. With the current code it is displayed very quickly before the reload and is then gone.

i will fix it

yeneastgate:
> The notice "The form was saved" (or the equivalent for lists) should still be displayed. With the current code it is displayed very quickly before the reload and is then gone.

I fix it.pls, review code again, thanks

js/onoffice-copycode.js
```diff
@@ -0,0 +1,3 @@
+jQuery(document).ready(function($){
+	new ClipboardJS('#button-copy');
```
jmaas-onoffice, line 2: This library seems to use an [obsolete method](https://github.com/zenorocha/clipboard.js/issues/683). [`navigator.clipboard`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard) has [good support](https://caniuse.com/mdn-api_navigator_clipboard), so could we use that instead?

js/admin.js
```diff
@@ -1,4 +1,13 @@
 jQuery(document).ready(function($){
+	if ( sessionStorage.getItem('message_save_view') ) {
```
jmaas-onoffice, line 2: I think it would be better to implement this on the server-side instead of in JS. There are some edge cases, like if I reload the page and resend the form, where this value will not be set on the client-side but the server will actually resave the form.

Could you implement this on the PHP-side that when the form is saved this notice is displayed?

jmaas-onoffice (changes requested):

css/admin.css
```diff
@@ -326,4 +326,22 @@ div[data-name^=oopluginfieldconfigformdefaultsvalues-value].multiselect > input.
...
+
+body #post-body-content .viewusage, body #post-body-content .wp-clearfix{
+	display: inline-block;
+}
+body #post-body-content p.wp-clearfix:nth-child(1){
+	display: block !important;
+}
+#button-copy{
+	margin-left: 5px;
+	font-size: 1.2em;
```
jmaas-onoffice, line 346: Also, looking at the [video](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/160#issuecomment-1135767892), the button for copying looks pretty large. It should be the same size as the other WordPress-buttons.

Could this be because of this `font-size`?

jmaas-onoffice (changes requested):

LongTrong-exe (dismissed):

js/clipboard/clipboard.min.js
```diff
@@ -0,0 +1,7 @@
+/*!
+ * clipboard.js v2.0.11
```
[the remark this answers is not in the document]
jmaas-onoffice: Do you have documentation on that? As far as I can tell from MDN, it should work as long as it's called in an event that was caused by the user.

"Writing to the clipboard is available without permission in secure contexts and browser extensions, but only from user-initiated event callbacks. Browser extensions with the "clipboardWrite" permission can write to the clipboard at any time." (https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText)

How much work is it to try it out and test it?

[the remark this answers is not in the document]
dai-eastgate: Please take a look this article: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard. 

> Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility).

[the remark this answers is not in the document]
jmaas-onoffice: Don't we only need [`writeText`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText)?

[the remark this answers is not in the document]
dai-eastgate: yes, but navigator.clipboard isn't available on insecure context. We can't access the writeText function.

js/admin.js
dai-eastgate: Maybe I misunderstand this comment. As I know we always save the message to the sessionStorage after submitting the form. The case you mention above can't happen. For the PHP-side solution, what is the steps to be done? After submitting the form, the page will be redirected to edit page. Maybe we need the query params to check if you decide to work on PHP-side.

dai-eastgate: @jmaas-onoffice did you take a look this one?

jmaas-onoffice: So right now this is the mechanism:

1. Press "Save".
2. Store that we just clicked "Save" in sessionStorage.
3. Send the form to WordPress.
4. WordPress returns the new edit page.
5. We check sessionStorage to see if we need to display the message.

But why do we need to do this in JavaScript? Can't we check in PHP whether we just saved and display the message directly in PHP, instead of using JavaScript?

So the mechanism would be:

1. Press "Save".
2. The form is sent to WordPress.
3. In PHP when we are saving the form, we also display the "List saved" message.

Here, no JavaScript is needed and we don't need to track button clicks, because in PHP we know whether we are saving the form or not.

Is this a bit clearer?

yeneastgate (approved):

yeneastgate (dismissed):

yeneastgate (dismissed):

dai-eastgate:
Hi @jmaas-onoffice, could you please review this PR? Thanks.

plugin/Gui/AdminPageSettingsBase.php
```diff
@@ -246,7 +246,25 @@ public function ajax_action()
 			$this->updateValues($row, $pResultObject, $recordId);
 		}
 
-		$pResultObject->messageKey = $this->getResponseMessagekey($pResultObject->result);
+		$pResultObject->messageKey = $this->getResponseMessagekey( $pResultObject->result );
+		if ( array_key_exists( 'oo_plugin_listviews', $row ) ) {
+			$pResultObject->shortcodeContent = '<span class="viewusage">' . esc_html__( 'Shortcode: ',
+					'onoffice-for-wp-websites' ) . '<input style="margin-right: 5px;" type="text" readonly="" value="[oo_estate view=&quot;' . $row['oo_plugin_listviews']['name'] . '&quot;]"></span>';
+			if ( wp_is_using_https() ) {
```
jmaas-onoffice, line 253: There is an even better way to do the progressive enhancement: You can check if you have access to the necessary functions in JS.

So you would always load a JS-script that would check if it can actually use `navigator.clipboard.writeText()`, and if it can, it will add the button. That means that if anything goes wrong (JS not loaded, or `navigator.clipboard` not available), the button will not be shown. It will only be present if it actually works.

Could you try that?

js/ajax_settings.js
```diff
@@ -43,6 +45,7 @@ onOffice.ajaxSaver = function(outerDiv) {
 			var newKey = mergeElement[i];
 			data[newKey] = onOffice.settings[newKey];
 		}
+		$('#onoffice-notice-wrapper').empty();
 
 		jQuery.post(onOffice.settings.ajax_url, data, function(response) {
```
jmaas-onoffice, line 50: I'm just noticing that we are still doing an AJAX post. The idea behind this PR is to not use the async JS POST, but the normal form post that is common for WordPress. This is what I meant with "reload edit pages after saving", sorry if that wasn't clear!

Is the difference clearer now or do you have questions? We can maybe do this change in another PR, and focus on the copy button in this one.

jmaas-onoffice (changes requested):

js/ajax_settings.js
dai-eastgate: I used the ajax call like creating a WordPress post. They will use the replaceState to change the url. I will update this to normal form post if you think it will be better.

yeneastgate (dismissed):

hungnc89:
- Tester retested
- Video:

https://user-images.githubusercontent.com/72601287/173515888-1af96d35-a704-4c67-9e9c-7f520b7bcd9a.mp4

plugin/Gui/Table/EstateListTable.php
```diff
@@ -226,9 +226,10 @@ public function get_columns()
 	 *
 	 */
 
-	protected function column_shortcode($pItem)
+	protected function column_shortcode( $pItem )
 	{
-		return '<input type="text" readonly value="[oo_estate view=&quot;'.esc_html($pItem->name).'&quot;]">';
+		return '<input type="text" style="max-width: 100%; margin-right: 5px;" readonly value="[oo_estate view=&quot;' . esc_html( $pItem->name ) . '&quot;]"><input type="button" class="button button-copy" data-clipboard-text="[oo_estate view=&quot;' . esc_html( $pItem->name ) . '&quot;]" value="' . esc_html__( 'Copy',
+				'onoffice-for-wp-websites' ) . '" >';
```
jmaas-onoffice, line 232: I just noticed that the button becomes visible slightly after the rest of the page has loaded. I didn't think of that. But there seems to be a solution: If you use an inline script right after the button and set the class there, it will happen before the first paint. So could you move the check whether the button should be shown from the script file into an internal script?

For example, you could change this line to:
```
return '<input type="text" style="max-width: 100%; margin-right: 5px;" readonly value="[oo_estate view=&quot;' . esc_html( $pItem->name ) . '&quot;]"><input type="button" class="button button-copy" data-clipboard-text="[oo_estate view=&quot;' . esc_html( $pItem->name ) . '&quot;]" value="' . esc_html__( 'Copy',
				'onoffice-for-wp-websites' ) . '" >
<script>if (navigator.clipboard) { jQuery('.button-copy').show(); }';</script>
```

You would need to make this change for all places where the button is rendered. You can then remove this check from the script (https://github.com/onOfficeGmbH/oo-wp-plugin/blob/44f9412b1c06f016dd9373564c90e8e9762658dc/js/onoffice-copycode.js).

Does this make sense?

jmaas-onoffice (changes requested):
It seems that address lists are still on the old behavior. Could you make them behave the same as the others?

plugin/Gui/Table/EstateListTable.php
dai-eastgate: Yes I will update this

dai-eastgate:
> It seems that address lists are still on the old behavior. Could you make them behave the same as the others?

Just for the address lists right? not for the Estate Detail, Similar Estates...

jmaas-onoffice:
Oh, you're right. It would be great to have everything behave the same. So please update all of those to the new behavior. Thanks!

dai-eastgate:
Currently, the address list and the address detail don't have the copy button. Do I need to update for it also? This issue you just mention the estate list & form.

jmaas-onoffice:
If it is not a lot of extra work, then yes, let's add the copy button everywhere. :)

LongTrong-exe (dismissed):

dai-eastgate:
This is the test video:

https://user-images.githubusercontent.com/106214469/175010540-6e80823b-b129-46bc-94af-15db25d150ce.mp4

jmaas-onoffice (approved):
There's too many changes for me to review them, but I will test it.

jmaas-onoffice (changes requested):
1. The detail view does not use the new shortcode display:
    
    grafik
    
    It should look like this:
    
    grafik

2. Similarly, when the detail view shortcode is not used
    grafik

    it should look like this:
    
    grafik

3. Unit list settings do not show a shortcode.

    grafik

    They should have the same shortcode as the estate list settings (also the same behavior when creating a new unit list).

dai-eastgate:
Do we apply the new display shortcode for other pages (estate list, address list, form, unit lits)?

LongTrong-exe (approved):

jmaas-onoffice (approved):
It all works great, thanks! Another PR finally finished. 🎉

Issue #161, opened 2022-01-04
P#62389 #2781085 address list without of a name created in WP-plugin


michaelarnold-dev:
Tested in v2.20.12-1-gb936a28e on https://wpptest.onofficeweb.com 

Addresses -> An address list without a name exists. How could it be created?

michaelarnold-dev:
@nglelinh : Do you have an idea how that can be fixed?

michaelarnold-dev:
@nglelinh Any update here?

nglelinh:
@michaelarnold-dev we did not check this issue yet

vu-vuong-esg:
@michaelarnold-dev , i re produce issue with video, is that issue you are referring to?

https://user-images.githubusercontent.com/72490437/149263979-a2120e20-cc58-43d0-bc74-e767b8fdfe17.mp4

michaelarnold-dev:
The tester says that the following video shows the issue:
https://user-images.githubusercontent.com/77264788/149886476-5c2c0ca3-ca3b-483e-81da-89572cd47b9a.mp4

vu-vuong-esg:
i will fix it

michaelarnold-dev:
@vu-vuong-esg Is the fixing process still in progress?

tang-hien-egs:
> @vu-vuong-esg Is the fixing process still in progress?

i fixed done, can you check PR https://github.com/onOfficeGmbH/oo-wp-plugin/pull/172

PR #172, opened 2022-01-20
P#62389 #2781085 Address list without of a name created

vu-vuong-esg:
P#62389 #2781085 Address list without of a name created
#161


jayay:
I don't quite understand the change?

vu-vuong-esg:
> I don't quite understand the change?

i update it, can you check PR ?

plugin/Record/RecordManagerUpdateListViewAddress.php
```diff
@@ -94,7 +94,10 @@ private function insertNewRecords(array $records, string $table): bool
 		$pWpDb = $this->getWpdb();
 		$result = true;
 		foreach ($records as $record) {
-			$result = $result && $pWpDb->insert($prefix.$table, $record);
+			if(is_array($record))
```
jayay, line 97: This is better. @vu-vuong-esg do you know, under which circumstances `$record` is not an array?

vu-vuong-esg: because $record is string so it doesn't know column in table insert.

vu-vuong-esg:
as in the chat, i add code in fixed error create address list can't save page.

plugin/Record/RecordManagerUpdateListViewAddress.php
jayay: In which case is `$record` a string? What is the root cause?

tang-hien-egs:
@jayay , i fixed it. can you check PR ?

jayay:
Isn't the origin of the bug in the way the JavaScript code submits the data to be saved?

vu-vuong-esg:
> Isn't the origin of the bug in the way the JavaScript code submits the data to be saved?

@jayay , No, JavaScript code submits data true

plugin/Record/RecordManagerUpdateListViewAddress.php
jayay: @vu-vuong-esg are you sure `$row` has the correct type here and doesn't need to be wrapped in another array?
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/f328ccafd4dc37a672ab9d33b156d00e071ac263/plugin%2FGui%2FAdminPageAddressListSettings.php#L233-L236

jayay (approved):

michaelarnold-dev:
@tang-hien-egs Could you try to increase the code coverage?

jayay (approved):

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
I can still reproduce the error with the [latest code](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/172#issuecomment-1094942595).

I have configured this version at https://3.wptest.onofficeweb.com/. If you create a new address list there without entering a name and save, then the list will actually still be created without a name.

Could you check this on the test instance please?

tang-hien-egs:
> I can still reproduce the error with the [latest code](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/172#issuecomment-1094942595).
> 
> I have configured this version at https://3.wptest.onofficeweb.com/. If you create a new address list there without entering a name and save, then the list will actually still be created without a name.
> 
> Could you check this on the test instance please?
@jmaas-onoffice 
this errors like link record : https://files.fm/f/gzgshnbxc , your mean is when created address list, we must enter name address not exist

jmaas-onoffice:
> this errors like link record : https://files.fm/f/gzgshnbxc , your mean is when created address list, we must enter name address not exist  ?

Yes, the video shows the error. With estate lists and forms, when you try to save without a name, it gives an error and you can't save. The same should happen for address lists.

Is the problem clear? :)

tang-hien-egs:
> > this errors like link record : https://files.fm/f/gzgshnbxc , your mean is when created address list, we must enter name address not exist  ?
> 
> Yes, the video shows the error. With estate lists and forms, when you try to save without a name, it gives an error and you can't save. The same should happen for address lists.
> 
> Is the problem clear? :)

yes, i will fixed it as soon as !

tang-hien-egs:
@jmaas-onoffice I have a issue I don't know if it is in your request or not ? This is, With estate lists, forms and address list, when you enter name new and you press "Save" 2x in a row, it created data success. See video in link : https://files.fm/f/q7c5wyrxu

jmaas-onoffice:
@tang-hien-egs Oh, thanks for reporting. That is a bit weird. But since the last changes will be saved, I think this is ok. We can fix that later.

vu-vuong-esg:
> @tang-hien-egs Oh, thanks for reporting. That is a bit weird. But since the last changes will be saved, I think this is ok. We can fix that later.
> 
> Is the code ready for review apart for this issue?

@jmaas-onoffice  yes, the code ready for review.

jmaas-onoffice (approved):

jmaas-onoffice (approved):

PR #183, opened 2022-02-21
#2882701 Group contact persons' data fields

jmaas-onoffice:
Since the Web department implemented this change itself, there is no special project attached.


onOffice-Web (approved):
Trigger new zip.

PR #186, opened 2022-02-24
P#63827 Add a default email address

hungnc89:


hungnc89:
@jayay This task has not been finished yet because it is depended on P#56183 [https://github.com/onOfficeGmbH/oo-wp-plugin/pull/139]. That PR is not merged.

jayay:
#139 was merged today 🎉

plugin/Gui/Table/WP/ListTable.php
```diff
@@ -167,6 +167,11 @@ protected function handleRecord(array $records)
 				$pages = implode(',',$listPage);
 				$record->page_shortcode = $pages;
 			}
+			if ($record->default_recipient){
+				$record->recipient = "Default (".get_option('onoffice-settings-default-email', '').")";
+			} else {
+				$record->recipient = "$record->recipient (override)";
```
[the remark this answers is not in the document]
hungnc89: @jayay Yes, this is correct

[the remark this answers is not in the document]
jayay: I think escaping is missing here

plugin/Renderer/InputModelRenderer.php
```diff
@@ -219,6 +219,7 @@ private function createInputField(InputModelBase $pInputModel, FormModel $pFormM
 			case InputModelOption::HTML_TYPE_EMAIL:
 				$pInstance = new InputFieldEmailRenderer('email', $elementName);
 				$pInstance->addAdditionalAttribute('size', '50');
+				$pInstance->addAdditionalAttribute('pattern', '^[\w.-]+@[\w.-]+\.[\w]{2,}$');
```
jayay, line 222: why not just `<input type="email">`?

plugin/Gui/Table/WP/ListTable.php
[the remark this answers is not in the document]
hungnc89: I see, I fixed it though

plugin/Renderer/InputModelRenderer.php
hungnc89: `type="email"` only check @ sign.  For example, 'test@gmail' is considered a valid email in that case. I added `pattern`  attribute to make it cover more invalid cases

jmaas-onoffice: Yes, email validation is suprisingly difficult.

I'm going to decide that we don't need `pattern` validation and `type="email"` is enough.

This is supported by https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#pattern_validation, where they say you only need pattern validation if you want the inputs to be 'restricted further than just "any string that looks like an e-mail address"'. This is not the case for us, we accept all email addresses here. If the user enters something invalid, we can't really do much about it.

Therefore, please remove the `pattern`. :)

jmaas-onoffice (changes requested):

plugin/Renderer/InputModelRenderer.php
hungnc89: ok, I removed it

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
1. The requirements didn't specify what the default setting for new forms is, sorry! Please implement that for newly created forms, the checkbox "Use default email address" is checked.
2. Could you remove the placeholder from the override input which is shown in this image? It should be empty.
    grafik
3. When the defaul email address is missing, in the overview there are empty parentheses and in the settings there is nothing. Could you add "missing" in parentheses and italic, as shown in these two screenshots?
  grafik
  grafik

plugin/Gui/AdminPageFormSettingsInquiry.php
```diff
@@ -52,7 +52,8 @@ protected function checkFixedValues($row)
 	{
 		$table = RecordManager::TABLENAME_FORMS;
 		$resultName = isset($row[$table]['name']) && $row[$table]['name'] != null;
-		$resultRecipient = isset($row[$table]['recipient']) && $row[$table]['recipient'] != null;
+        $resultRecipient = ($row[$table]['default_recipient'] == 0 && isset($row[$table]['recipient']) && $row[$table]['recipient'] != null)
```
LongTrong-exe, line 55: Please re-align and use tabs instead of spaces

plugin/Gui/Table/FormsTable.php
```diff
@@ -205,6 +206,17 @@ protected function column_shortcode($pItem)
 			.esc_html($pItem->name).'&quot;]">';
 	}
 
+    protected function column_recipient($pItem)
```
LongTrong-exe, line 209: here too

plugin/Gui/Table/WP/ListTable.php
```diff
@@ -167,6 +167,15 @@ protected function handleRecord(array $records)
 				$pages = implode(',',$listPage);
 				$record->page_shortcode = $pages;
 			}
+			if ($record->default_recipient){
+                if (get_option('onoffice-settings-default-email', '')) {
```
LongTrong-exe, line 171: And here

plugin/Model/FormModelBuilder/FormModelBuilderDBForm.php
```diff
@@ -264,12 +264,41 @@ public function createInputModelEmbedCode()
...
 	}
 
+	/**
+	 * @return InputModelDB
+	 * @throws Exception
+	 */
+	public function createInputModelDefaultRecipient(): InputModelDB
+	{
+		$isDefaultEmailMissing = false;
+        if (get_option('onoffice-settings-default-email', '') !== '') {
```
LongTrong-exe, line 274: Here

LongTrong-exe (changes requested):
Please re-align and use tabs instead of spaces

plugin/Gui/Table/FormsTable.php
```diff
@@ -206,6 +206,17 @@ protected function column_shortcode($pItem)
...
+    {
+        $emailStatusMissing = 'missing';
+        $pos = strpos($pItem->recipient, $emailStatusMissing);
```
LongTrong-exe, line 212: Could name of the variable `$pos` be more specific?

LongTrong-exe (changes requested):

plugin/Renderer/InputFieldItalicLabelCheckboxRenderer.php
```diff
@@ -0,0 +1,69 @@
+<?php
```
jmaas-onoffice, line 1: From the code I'm not sure I understand. What problem does this class solve?

jmaas-onoffice (changes requested):
Could you merge `master` into this PR and resolve the conflicts?

plugin/Renderer/InputFieldItalicLabelCheckboxRenderer.php
hungnc89: This class is for this kind of label which mix normal font with italic font
image

jmaas-onoffice: Why do we need a separate class for this? Is it because the other class only takes text and this needs HTML?

LongTrong-exe: You're right

LongTrong-exe: You're right

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
It appears that one of the requirements doesn't work anymore.

On https://1.wptest.onofficeweb.com/wp-admin/admin.php?page=onoffice-forms you can see that the I set a contact form to the default email address, but the default email address is missing. However, no warning appears.

grafik

For convenience, here is the requirement for the warning:

> 4) Add a notice for when the default email address is missing.
> 4.1) The notice is only shown when the default email setting is empty and when at least one form uses the default email. (That means that if all forms override the email, the default email setting may be empty without notice.)
> 4.2) The notice is shown on all pages of the backend. It cannot be dismissed until the issue is fixed.
> 4.3) The notice should have the following content:
> """
> The onOffice plugin is missing a default email address. Forms that use it will not send emails. Please add a default email address in the <a href="<link to settings">plugin settings</a>.
> """

LongTrong-exe:
@jmaas-onoffice This error occurs when the user updates to this version, the "default email" option has not been created, it will use the default value, not null, causing the warning to not appear.
I think this error appeared before.

jmaas-onoffice (changes requested):
Does the line https://github.com/onOfficeGmbH/oo-wp-plugin/pull/186/files#diff-d7e45a4f7e97aabdaea156cc6b687822f80cdb610a3a1c6d92dc70188605695aR407 also need to be changed?

LongTrong-exe:
> Does the line https://github.com/onOfficeGmbH/oo-wp-plugin/pull/186/files#diff-d7e45a4f7e97aabdaea156cc6b687822f80cdb610a3a1c6d92dc70188605695aR407 also need to be changed?

I think not, because I added that option in `DatabaseChange.php` when user update to new version, so it always exists
https://github.com/onOfficeGmbH/oo-wp-plugin/blob/f9ca3c530d6fd7dd84a2722387d9a3de5e42b84b/plugin/Installer/DatabaseChanges.php#L194

jmaas-onoffice:
Oh, the link is broken. I hope we're talking about the same line... 😅 https://github.com/onOfficeGmbH/oo-wp-plugin/blob/bf201f6ba44dc6dfa63763f96c30ce84292108d6/plugin/Controller/AdminViewController.php#L407

Why is the default for this option set to `'onoffice-for-wp-websites'`? I thought that value is the text domain, so it does not make sense to me in this context.

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
Looks good!

I'd like to make a minor visual improvement. When I open the settings of a form, it takes a little while to load. In that time, the "Override email address" is active. Once loaded it gets deactivated.

I think it's more natural to flip it around: During load have the "Override email address" deactivated and after load, if it should be active, activate it.

Does this make sense?

LongTrong-exe:
@jmaas-onoffice I also noticed what you said before, let me fix it.

LongTrong-exe:
https://user-images.githubusercontent.com/101325858/169436669-99e3a921-7a9d-4084-a4ee-3a6292206b8d.mp4

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
Sorry, I just noticed that I forgot to put a requirement:

New forms should use the default email. Currently new forms use the override.

Could you change this? Everthing else already works great! :)

jmaas-onoffice (dismissed):

jmaas-onoffice:
@LongTrong-exe Could you fix the conflicts? I'd like to try and add this to the next release that I'm aiming to release next week. :)

jmaas-onoffice:
Thanks! I'll do a round of testing to prepare the release now. :)

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
1. The applicant search form shows `(override)` as an email address.
    grafik
    It can't have an email address, so it should show a dash (`-`) to indicate that this field is not set.
2. The strings for the column are not translatable: https://github.com/onOfficeGmbH/oo-wp-plugin/blob/18b371aa32c9fed674319cecd2446be0e922b371/plugin/Gui/Table/WP/ListTable.php#L172-L177
    Please make them translatable.

jmaas-onoffice (changes requested):
1. The applicant search form shows `(override)` as an email address.
    grafik
    It can't have an email address, so it should show a dash (`-`) to indicate that this field is not set.
2. The strings for the column are not translatable: https://github.com/onOfficeGmbH/oo-wp-plugin/blob/18b371aa32c9fed674319cecd2446be0e922b371/plugin/Gui/Table/WP/ListTable.php#L172-L177
    Please make them translatable.

jmaas-onoffice (dismissed):

jmaas-onoffice:
I just merged some other PRs. Could you update this branch and resolve the conflicts?

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
1. For the contact form, the email override starts out deactivated as desired. For interest and owner forms, it starts out activated and when the default email is active, it suddenly gets disabled after loading. Please make it consistent with the contact form for all form types.
2. In the contact form, the label for the email is "Override email address", but for the other form types it is "Override Recipient's E-Mail Address". Please make it consistent and use "Override email address" for all form types.

LongTrong-exe (dismissed):

LongTrong-exe:
@jmaas-onoffice On master branch, email input in all forms has type 'text', so I update it to 'email' type and add email validation when saving form.

jmaas-onoffice:
@LongTrong-exe Yes, that sounds good, thanks.

I think we had some discussion on how to validate email addresses in another PR, but I can't find it. The conclusion was: Don't do anything complicated, `type="email"` is enough. ;)

jmaas-onoffice:
Oh, I didn't see that you already requested my review. I will have a look!

js/ajax_settings.js
```diff
@@ -11,9 +11,24 @@ onOffice.ajaxSaver = function(outerDiv) {
...
 	this.register = function() {
 		var proto = this;
-		this._outerDiv.find('#send_ajax').on('click', function() {
-			window.scrollTo(0, 0);
-			proto.save();
+		var form = $("form");
+		var typesNeedValid = ['email'];
+
+		form.submit(function (e) {
+			e.preventDefault();
```
jmaas-onoffice, line 18: I'm a bit confused by this. Why do we need to prevent the form from sending and validate and save it ourselves here? Shouldn't that be something we do in #160?

jmaas-onoffice (changes requested):

js/ajax_settings.js
LongTrong-exe: I think it would be better to fix that in #160

jmaas-onoffice (approved):

PR #187, opened 2022-03-08
P#63829 Edit pages load too long and break when many regions are configured i…

tang-hien-egs:
P#63829 Edit pages load too long and break when many regions are configured i…


plugin/Field/Collection/FieldLoaderEstateRegionValues.php
```diff
@@ -0,0 +1,122 @@
...
+		$this->_pSDKWrapper = $pSDKWrapper;
+		$this->_pRegionController = $pRegionController;
+		$this->_pRegionFilter = $pRegionFilter;
+	}
+
+	/**
+	 * @return Generator
+	 * @throws APIEmptyResultException
+	 */
+	public function load(): Generator
```
jayay, line 73: This does the same as `FieldLoaderGeneric` and adds the region loading on top. I think, this should either _only load the regions_ or _use `FieldLoaderGeneric`_ so that the code isn't duplicated

jayay: It'd be nice if there was a test for the new class as well.

jayay (changes requested):

jayay (approved):

jmaas-onoffice (approved):
Trigger new zip.

jmaas-onoffice (approved):
Trigger new zip.

jmaas-onoffice (approved):
Trigger new zip.

PR #190, opened 2022-03-17
#2932899 Fix detail view being inaccessible for reference estates

jmaas-onoffice:


jmaas-onoffice:
What's a better name for the test method?

jayay:
Best case would be to create two test methods:

`testReferenceStatusForDetailPageFalse` - asserts  that reference status is false for detail pages
`testFalseReferenceStatusGeneratesEmptyMarketingStatus` - asserts  that a false reference status leads to an empty marketing status (for an estate that would otherwise have data that leads to a non-empty marketing status)

jayay (dismissed):

jayay (approved):

PR #191, opened 2022-03-28
Development build documentation

jmaas-onoffice:
Document how to do a development build and improve the scripts.


jmaas-onoffice:
Let us test that the changes to the build script work by giving an approving review. ;)

.github/workflows/dev-build.yml
```diff
@@ -35,24 +35,19 @@ jobs:
       run: composer install
 
     - name: Build for development
-      run: PREFIX=/tmp/release make test-zip
```
jayay, line 38: `make release` names the plugin _onOffice for WP-Websites_ which can lead to the issue, that a WP instance that has the pre-release version of the plugin installed, can update itself and overwrite it with the version from wordpress.org. This is the reason why there is the `make test-zip` target: it holds on the the plugin name _onOffice for WP-Websites *(dev)*_. `make test-zip` definitely needs to stay.

.github/workflows/dev-build.yml
```diff
@@ -35,24 +35,19 @@ jobs:
...
-    - run: cat /tmp/release/onoffice/plugin.php
-
     - name: Zip files
       run: (cd /tmp/release/ && zip -r ./onoffice-${{ env.DEV_BUILD_VERSION}}.zip ./onoffice/)  # To prevent nested paths (like /tmp/release/onoffice/readme.txt) we have to cd into the directory.
 
     - name: Upload zip
       uses: actions/upload-artifact@v3
       with:
         name: onoffice-${{ env.DEV_BUILD_VERSION }}-please-unpack
-        path: /tmp/release/onoffice-${{ env.DEV_BUILD_VERSION }}.zip  # Uploading a .zip leads to a .zip inside a .zip, but uploading the folder takes significantly longer.
```
jayay, line 55: the old version was fine

jayay (changes requested):

BUILDING.md
```diff
@@ -0,0 +1,10 @@
+# Building
+
+This is how you can generate a .zip to upload to a WordPress instance.
+
+1. Run `composer install`.
+    - If it asks you to update, run `composer update`.
```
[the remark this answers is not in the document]
jmaas-onoffice: Thanks for the explanation!

.github/workflows/dev-build.yml
jmaas-onoffice: Oh, I overlooked that release changes the name. Then we'll simply .zip it twice.

.github/workflows/dev-build.yml
jmaas-onoffice: Do you mean the comment? I'm not sure it is self-explanatory, so I tried to provied a bit more context.

BUILDING.md
```diff
@@ -0,0 +1,9 @@
+# Building
+
+This is how you can generate a .zip to upload to a WordPress instance.
+
+1. Run `composer install`.
```
jayay, line 5: I would prefer something like
__You'll need the PHP extensions `mbstring`, `curl` and `simplexml`. If you're using Ubuntu, you can install these by running `sudo apt install php-mbstring php-curl php-simplexml`__ just to not pretend Ubuntu is the only operating system out there.

jayay: Also, running `apt install` for a package that's already installed doesn't hurt, so there is not really a reason not to have them all in one line

jayay (commented):
I commented on some things. Nothing mandatory, just thoughts.

BUILDING.md
jmaas-onoffice: How can we check which extensions are required?

onOffice-Web (approved):
Test approval for action.

onOffice-Web (approved):
Test approval for action.

onOffice-Web (approved):
Test approval for action.

onOffice-Web (approved):
Test approval for action.

PR #194, opened 2022-04-01
Rebuild release actions

jmaas-onoffice:
- Adds a "Stable tag: <current version>" to readme.txt.
- Uses the official title in plugin.php. (No suffix "(dev)".)
- Adds the current stable version to plugin.php.
- The new development zip installs the plugin into the same folder as the released plugin (/plugins/onoffice-for-wp-websites). But it is distinguishable from the release by its version (it has the commit hash from `git describe --tags`).

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -1,0 +1,69 @@
+name: Build Release
+
+on:
+  workflow_call:
+    inputs: 
+      ref:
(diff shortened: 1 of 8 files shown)


onOffice-Web (dismissed):
Test review approval to see if it triggers the new action.

onOffice-Web (dismissed):
Another action test.

onOffice-Web (dismissed):
Another action test.

onOffice-Web (dismissed):

PR #200, opened 2022-04-07
P#54045 Insert custom variable yoast seo

yennt89:
https://www.loom.com/share/0ee7d1ed548b40b98e35a667c07fef8a


plugin.php
```diff
@@ -135,6 +135,39 @@
 	return $pDI->get(EstateViewDocumentTitleBuilder::class)->buildDocumentTitle($title);
 }, 10, 2);
 
+if (in_array('wordpress-seo/wp-seo.php', apply_filters('active_plugins', get_option('active_plugins')))) {
```
jmaas-onoffice, line 138: This would only work for this specific plugin. People may choose another plugin and then it would break.

The appraoch with the custom fields works for all SEO plugins that I have found. What was the issue with that approach?

yennt89: Yes, I will try it with another plugin. and research solutions for this issue

yennt89: I tried it with another plugin (ex: Rank Math SEO)
I see that:
Each plugin will have a private hook to custom variables
ex:
Seo Yoast: wpseo_register_extra_replacements
Rank Math SEO: rank_math/vars/register_extra_replacements

I pushed the code and you can see it in this video :
https://www.loom.com/share/ca725ade1d354f7ab9c70b3d931d5cc1

So we can only support this feature for a specific list of plugins.

Please provide plugins, Which you want to implement this feature, or give me your opinion on this isssue.

jmaas-onoffice: I have found a way to get the estate ID.

For testing, I inserted the following code into my theme's `functions.php`:
```php
function add_oo_meta($value, $object_id, $meta_key) {]
    $detail_page_id = 148;
    if ($object_id == $detail_page_id && $meta_key == "oo_estate_id") {
        return print_r(get_query_var("estate_id"), true);
    } else {
        return null;
    }
}
add_filter("get_post_metadata", "add_oo_meta", 1, 3);
```

I can then add the custom field `oo_estate_id` in Yoast:

grafik

This allows me to put the estate ID into the title. In a sense, it goes from the URL to the query var to the custom field to the title.

grafik

If you can access the estate ID in the hook for the custom field, can you then get the desired information about the estate?

yennt89: Yes, I will try it.

yennt89: @jmaas-onoffice I resolved this issue.
Pls, review the code again.
This video test:
https://www.loom.com/share/374bd37519384d2a9ea7240d4b3ea3a8

plugin.php
```diff
@@ -135,6 +135,32 @@
 	return $pDI->get(EstateViewDocumentTitleBuilder::class)->buildDocumentTitle($title);
 }, 10, 2);
 
+add_filter('get_post_metadata', function($value, $object_id, $meta_key) use ($pDI) {
+	$pDataDetailViewHandler = $pDI->get(DataDetailViewHandler::class);
+	$pDetailView = $pDataDetailViewHandler->getDetailView();
+	$detail_page_id = $pDetailView->getPageId();
+	if ($object_id == $detail_page_id) {
+		$list_meta_keys = ["onoffice_titel" => '%1$s', "onoffice_title" => '%1$s',
```
jmaas-onoffice, line 143: The way this is currently, the `%1$s`, `%2$s`, ... are magic numbers that are difficult to understand.

Since you format them all as strings anyway `$s`, what do you think of using this array to map the custom field key to the enterprise field key? So for example `"onoffice_titel" => 'objekttitel'.

Then you don't need to look in another part of the code what the different numbers mean. ;)

plugin/Controller/EstateTitleBuilder.php
```diff
@@ -109,6 +110,44 @@ public function buildTitle(int $estateId, string $format): string
...
+	 * %7$s: 'Id'
+	 *
+	 * @return string
+	 * @throws Exception
+	 */
+
+	public function buildCustomFieldTitle(int $estateId, string $format): string
+	{
+		$this->_pDefaultFilterBuilder->setEstateId($estateId);
+		$this->_pEstateDetail->loadSingleEstate($estateId);
```
jmaas-onoffice, line 134: Does this mean that we send two requests to enterprise?

plugin/ViewFieldModifier/EstateViewFieldModifierTypeTitle.php
```diff
@@ -65,6 +65,27 @@ public function getAPIFields(): array
...
+			'objekttitel',
+			'objektbeschreibung',
+			'ort',
+			'plz',
+			'objektart',
+			'vermarktungsart',
+			'Id'
+		];
+
+		return array_values(array_unique(array_merge($this->_viewFields, $titleFields)));
```
jmaas-onoffice, line 86: Since the custom fields can only use the fields from `$titleFields`, why are we merging them with other fields?

jmaas-onoffice (changes requested):
Very nice! From your video it seems that this appraoch works!

I have a few questions about the code organization. ;)

plugin.php
yennt89: Yes, I will research other solutions to solve it.

plugin/Controller/EstateTitleBuilder.php
yennt89: Yes, I will send two requests

plugin/ViewFieldModifier/EstateViewFieldModifierTypeTitle.php
yennt89: Because if it's not in the $this->_viewFields section it will not display on the front-end

plugin/Controller/EstateTitleBuilder.php
jmaas-onoffice: Ok, I'll have to check the performance. The double request will only be sent if the option to set custom fields is active, correct?

plugin/ViewFieldModifier/EstateViewFieldModifierTypeTitle.php
jmaas-onoffice: So the goal is if `$this->_viewFields` does not contain `plz` (because the user does not want to display it), but `$titleFields` contains it, then `plz` should _not_ be returned?

yennt89: Yes, that's right

yennt89: image

yennt89: image

jmaas-onoffice: Doesn't `array_merge` put both together? So if `$this->_viewFields = ["plz", "gesamtflaeche"]` and `$titleFields = ["plz", "objekttitel"]`, then doesn't `array_values(array_unique(array_merge($this->_viewFields, $titleFields)))` return `["plz", "gesamtflaeche", "objekttitel"]`, but you want `["plz"]`?

It seems to work, so I don't want to be too nit-picky about this, but I'm just surprised that it works. :)

yennt89: yes, i will check it again

LongTrong-exe (approved):

hungnc89:
- Tester  tested
- Video 

https://user-images.githubusercontent.com/72601287/164611254-aac65da0-a438-4586-9a88-b035f4c1ddbf.mp4

plugin/ViewFieldModifier/EstateViewFieldModifierTypeTitle.php
yennt89: @jmaas-onoffice I refactored the code, pls review it again, thanks

jmaas-onoffice (approved):

jmaas-onoffice (approved):

jmaas-onoffice:
I couldn't find the desired settings when I tested it. I have copied the requirements into GitHub at https://github.com/onOfficeGmbH/oo-wp-plugin/issues/223.

1. Could you please have a look at https://github.com/onOfficeGmbH/oo-wp-plugin/issues/223? If there are any questions, feel free to ask.
2. What about https://github.com/onOfficeGmbH/oo-wp-plugin/pull/118? We only need one PR, right? Should we close one?

jmaas-onoffice (changes requested):
I couldn't find the desired settings when I tested it. I have copied the requirements into GitHub at https://github.com/onOfficeGmbH/oo-wp-plugin/issues/223.

1. Could you please have a look at https://github.com/onOfficeGmbH/oo-wp-plugin/issues/223? If there are any questions, feel free to ask.
2. What about https://github.com/onOfficeGmbH/oo-wp-plugin/pull/118? We only need one PR, right? Should we close one?

yennt89:
> 

1.yes, I will check the requirements into GitHub at #223.
2.this pull request #200, I only implement the demo feature "Insert custom variable yoast seo" . we will close it and only implement at #118

yeneastgate:
@jmaas-onoffice All code implemented at #118. I will close this pull request

PR #217, opened 2022-04-19
P#69791 Migrate show reference estate setting properly

hungnc89:


js/onoffice-reference-estate-select.js
```diff
@@ -2,6 +2,12 @@ jQuery(document).ready(function() {
     const valueSelect = 'reference';
```
LongTrong-exe, line 2: I think name of the $valueSelect const should be more precise

plugin/Installer/DatabaseChanges.php
```diff
@@ -747,4 +748,14 @@ public function setDataDetailViewAccessControlValue()
 		$pDetailView->setHasDetailView(true);
 		$pDataDetailViewHandler->saveDetailView($pDetailView);
 	}
+
+	public function updateShowReferenceEstate()
```
LongTrong-exe, line 752: This function name should also be more precise

LongTrong-exe (dismissed):

hungnc89:
- Tester tested
- Video

https://user-images.githubusercontent.com/72601287/165700624-5e6a7423-32b6-4646-bc95-4102d52a19a6.mp4

jmaas-onoffice (approved):

plugin/Installer/DatabaseChanges.php
```diff
@@ -781,4 +782,14 @@ public function setDetailTemplate()
 		$pDetailView->setTemplate( key( $firstTemplatePath ) );
 		$pDataDetailViewHandler->saveDetailView( $pDetailView );
 	}
+
+	public function updateShowReferenceEstateOfReferenceList()
+	{
+		$prefix = $this->getPrefix();
+		$sql = "UPDATE {$prefix}oo_plugin_listviews
+				SET `show_reference_estate` = 1
+				WHERE `list_type` = 'reference'";
```
jmaas-onoffice, line 791: This was a bit ambiguous in the requirements, but this should be set for all list types.

The reason is that previously, reference estates were always shown, for all list types. So for backwards compatibility, we need to keep showing reference estates.

jmaas-onoffice (changes requested):

jmaas-onoffice (approved):

jmaas-onoffice (approved):

PR #218, opened 2022-04-19
P#69855 Fix duplicate check warning

hungnc89:


plugin/Controller/AdminViewController.php
```diff
@@ -311,17 +311,15 @@ public function enqueue_css()
 
 	public function enqueueExtraJs($hook)
 	{
-		if (__String::getNew($hook)->contains('onoffice')) {
```
LongTrong-exe, line 314: Removing the if condition here will result in `$pObject->doExtraEnqueues();` will be called on every page.

LongTrong-exe (commented):
According to the code, requirment 1 has not been fixed.

plugin/Controller/AdminViewController.php
hungnc89: That's right, cause the warning is appeared on every pages.

LongTrong-exe (approved):

jmaas-onoffice (approved):

hungnc89:
- Tester tested
- Video:

https://user-images.githubusercontent.com/72601287/168030730-2dce8fb1-c54b-4285-84b1-f75ed94a8d49.mp4

PR #211, opened 2022-04-12
P#65163 Detail page leads to error when it has a parent page (new PR after revert)

jmaas-onoffice:
I merged #182, but unfortunately I then noticed a bug with WPML that leads to an infinite redirect loop. So I have to reopen this PR. Sorry for the confusion.

The bug is that when I am on the detail page and I switch the language, there is an infinite redirect loop. You can test this on https://2.wptest.onofficeweb.com/, I have installed WPML there.

PR #182, opened 2022-02-18
P#65163 Detail page leads to error when it has a parent page

hungnc89:

PR #234, opened 2022-05-03
P#65163 Detail page leads to error when it has a parent page (new PR after revert, number 2)

onOffice-Web:
Previous PR: #211

Again, it all looked good, but after merging I found a bug, so I had to reopen the PR.


LongTrong-exe:
@jmaas-onoffice Have u investigated anything yet?

plugin/EstateList.php
```diff
@@ -398,6 +404,17 @@ public function estateIterator($modifier = EstateViewFieldModifierTypes::MODIFIE
 			$this->_currentEstate['id'];
 		$this->_currentEstate['title'] = $currentRecord['elements']['objekttitel'] ?? '';
 
+		if ( $this->_pDataView instanceof DataDetailView ) {
+			ob_start();
```
LongTrong-exe, line 408: I think if there's a bug, it will come from here. I can use `echo("<script>location.href = '".$url."'</script>");` as a safer solution instead of `wp_redirect`, what do u think?

LongTrong-exe: @jmaas-onoffice

jmaas-onoffice: The bug is that on one instance after upgrading, the detail page is broken. I need to save the detail page to make it work again. It might be related to the Enfold theme, so I'm trying to set up a test instance with it.

I will then see if the bug is related to what you are talking about here. Thanks for the hint!

jmaas-onoffice (approved):

plugin/Installer/DatabaseChanges.php
```diff
@@ -781,4 +793,27 @@ public function setDetailTemplate()
...
+	 *
+	 * @return void
+	 *
+	 */
+
+	public function checkAllPageIdsHaveDetailShortCode()
+	{
+		$pDataDetailViewHandler = $this->_pContainer->get(DataDetailViewHandler::class);
+		$pDetailView            = $pDataDetailViewHandler->getDetailView();
+		$shortCode              = '[oo_estate view="' . $pDetailView->getName() . '"]';
```
jmaas-onoffice, line 808: The shortcode does not always have the form `[oo_estate view="detail"]`. Other variants include:

- `[oo_estate    view="detail"]` (multiple spaces)
- `[oo_estate units="Units" view="detail"]` (with unit lists)

These are not detected, so after the update, the rewrite rule is missing and the detail page inaccessible.

You should change this piece of code to be more general so that any way of writing the shortcode for the detail view is recognized.

jmaas-onoffice (changes requested):
With the help of some colleagues I was able to find the problem. (It actually has nothing to do with Enfold, I just misremembered.)

jmaas-onoffice:
@LongTrong-exe Are you continuing the work on this? :)

LongTrong-exe:
> @LongTrong-exe Are you continuing the work on this? :)

Yes, why are you asking, what's the problem?

plugin/Installer/DatabaseChanges.php
LongTrong-exe: @jmaas-onoffice Oh, sorry, I forgot about this case, I hardly ever use those shortcodes.

jmaas-onoffice: We were lucky to have a test instance where this was the case. I also didn't think of this before! :)

jmaas-onoffice:
> > @LongTrong-exe Are you continuing the work on this? :)
> 
> Yes, why are you asking, what's the problem?

Oh, no problem, sorry! I just wasn't sure if I set everything up correctly or if you needed something else.

Sorry, if I caused confusion!

LongTrong-exe:
@jmaas-onoffice I have a question, can user use wp post for detail view?

jmaas-onoffice:
> @jmaas-onoffice I have a question, can user use wp post for detail view?

What's the problem? :)

LongTrong-exe:
> > @jmaas-onoffice I have a question, can user use wp post for detail view?
> 
> What's the problem? :)

Nothing, I asked because I don't know :sweat_smile:

jmaas-onoffice:
Sorry, I meant that I need more details to understand what you need to know. For what decision is it important to know, whether the shortcode can also be used in posts?

LongTrong-exe:
@jmaas-onoffice Is this a bug, or a feature? Because I've seen it exist for a long time. It also appears in "Similar Estates."
image

LongTrong-exe:
> > > @jmaas-onoffice I have a question, can user use wp post for detail view?
> > 
> > 
> > What's the problem? :)
> 
> Nothing, I asked because I don't know 😅



> Sorry, I meant that I need more details to understand what you need to know. For what decision is it important to know, whether the shortcode can also be used in posts?

When I use wp post for detail view. "estateId" is always removed in URL. I think it happens because of WP's Permalink Settings, but I don't know how to make it right.

jmaas-onoffice:
1. You're right, the [second breadcrumb](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/234#issuecomment-1120956080) is unnecessary. How much time would it take you to fix this?
2. Ok, the [shortcode in post](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/234#issuecomment-1120961447) is a weird behavior from WordPress. I have noted this in our internal bug tracker. But let's ignore this for now and assume that the shortcode is always placed in a page and never in posts.

jmaas-onoffice (approved):

LongTrong-exe:
@jmaas-onoffice Now it will look like this:

image
image

LongTrong-exe:
Now it will look like this:

image
image

jmaas-onoffice (approved):

jmaas-onoffice:
Thanks for noticing the issue with the double breadcrumbs! :)

jmaas-onoffice (changes requested):
Unfortunately I found more bugs...

I installed version [3.1-60-g8a07b32c](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/234#issuecomment-1126085304) on https://3.wptest.onofficeweb.com/ so you can test it there.

The issues appear when I select the theme "Twenty Twenty-Two". With other themes like "Neve" everything behaves as expected.

On version 3.1, the detail page looks like this:

grafik

After upgrading, the detail page seems to be missing header and side margins.

grafik

Another bug is that when I delete a WordPress page, I get a white page with only this error message:

```
Warning: Invalid argument supplied for foreach() in /www/htdocs/w01cc279/3.wptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Controller/DetailViewPostSaveController.php on line 141

Warning: Cannot modify header information - headers already sent by (output started at /www/htdocs/w01cc279/3.wptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Controller/DetailViewPostSaveController.php:141) in /www/htdocs/w01cc279/3.wptest.onofficeweb.com/htdocs/wp-includes/pluggable.php on line 1355

Warning: Cannot modify header information - headers already sent by (output started at /www/htdocs/w01cc279/3.wptest.onofficeweb.com/htdocs/wp-content/plugins/onoffice-for-wp-websites/plugin/Controller/DetailViewPostSaveController.php:141) in /www/htdocs/w01cc279/3.wptest.onofficeweb.com/htdocs/wp-includes/pluggable.php on line 1358
```

Could you have a look at what is causing the incompatibility with the theme and fix the issues?

LongTrong-exe:
@jmaas-onoffice Yeah, I checked, it seems to come from the error I mentioned here:
https://github.com/onOfficeGmbH/oo-wp-plugin/pull/234#discussion_r866457403
Using js to redirect is the safest and best way, but it makes the redirection not so smooth. 
And the 2nd error came from my mistake, I didn't see it before.
Thank you for reporting.

jmaas-onoffice:
The problem with a JS redirect is that search engines might not know about it, correct?

Could you investigate the issue for a few hours and see if you can identify the problem?

LongTrong-exe:
@jmaas-onoffice Using JS will make the page reload twice, you can watch the video:


https://user-images.githubusercontent.com/101325858/168722014-ef8f00f9-cb9d-4171-902d-df4138b676d1.mp4


About the problem with wp_redirect, you can read it [here](https://stackoverflow.com/a/8028987).

jmaas-onoffice:
The JS redirect is problematic and not a solution in our case. So let's try to find a different approach.

Why does it work for most themes but not the new page builder theme? What line of code is causing the issue?

LongTrong-exe:
@jmaas-onoffice 
It would be quite confusing to let me explain, so you can see the more specific explanations here:
https://stackoverflow.com/a/31366722
https://wordpress.stackexchange.com/a/81573
https://wordpress.stackexchange.com/a/76993

jmaas-onoffice:
I understand the problem with `wp_redirect`. But that's why you are using `ob_start()`, right?

And the error message "Warning: Cannot modify header information - headers already sent" [does not appear](https://github.com/onOfficeGmbH/oo-wp-plugin/pull/234#pullrequestreview-972401129) with the theme. So here it's not a problem with calling `wp_redirect` too late but something else, right?

LongTrong-exe:
> Cannot modify header information

Ever since the "Cannot modify header information" error appeared, it has only happened to some themes. https://github.com/onOfficeGmbH/oo-wp-plugin/pull/182#pullrequestreview-935051390

LongTrong-exe:
I refer to some documents, `ob_start()` will help fix it, but it is only a temporary solution, and leads to an interface error as you report.

I'm trying to run the `wp_redirect` function before the page's content is displayed, but if I do that I won't get the pageId (which will help to generate the correct url). So I'm still looking for a new, more efficient solution.

jmaas-onoffice:
I understand that for a good solution we need to find a different appraoch and that this will take time. That's ok, it's important to solve this well.

Do you think https://developer.wordpress.org/reference/hooks/template_redirect/ might help? We can later maybe also think about caching the API call, so that we don't send the same API call twice.

LongTrong-exe:
@jmaas-onoffice Can the user choose "Plain" type of Permalink?

image

jmaas-onoffice:
Why are you asking? What's the problem you need to solve? :)

LongTrong-exe:
> Why are you asking? What's the problem you need to solve? :)

I want to make sure, because when I select "plain" type, I won't be able to access the detail page.|

image

jmaas-onoffice:
Oh, ok. That's tricky.

Until now we don't have a lot of problems with that, so let's keep it like this for now. Meaning you can assume that the permalinks are set to "Post name". (I will have a look at how we can solve this, but it shouldn't block this.)

Thanks for reporting this problem!

LongTrong-exe:
@jmaas-onoffice I'm trying a new solution. I'll just change the "title" correctly and keep the rest of the link the same. With that, we will not need to use `pageId` and can use `wp_redirect` early.
The old solution uses `pageId` to get the page url and build a new link with the correct title.

What do you think?

jmaas-onoffice:
Sounds interesting, could you go into more detail?

So right now we get the page id and estate id from the URL via WP_Query, right? We use the page id to get the detail page's canonical URL. Then we use the estate id to get the estate's title and rebuild the new URL by appending the title to the canonical URL.

What do you mean when you say you will just change the title?

LongTrong-exe:
I only use estateId to get the title, keep the other path of the URL the same, then rebuild the new URL from them.

jmaas-onoffice (approved):

jmaas-onoffice:
You're right! This looks really good!

I will test this some more, but I think this might be the solution we will be going with. Thank you!

jmaas-onoffice (dismissed):

jmaas-onoffice:
I tested it and it works great. :)

One final change: When you make a fresh install of the plugin, the setting "Show title in URL" is deactivated by default. With this PR, it works well and should be active for new installations by default. Could you change that?

jmaas-onoffice (approved):

jmaas-onoffice:
We did it! 🎉

PR #253, opened 2022-05-30
P#51853 Add PHP 8 to unit tests

LongTrong-exe:


LongTrong-exe:
@jmaas-onoffice Needed phpunit 9. to support coverage so I updated it too

.github/workflows/unit-tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest]
-        php-versions: ['7.2','7.3', '7.4']
+        php-versions: ['7.3', '7.4', '8.0']
```
jmaas-onoffice, line 19: Our [official minimal PHP version](https://de.wordpress.org/plugins/onoffice-for-wp-websites/) for the plugin is 7.2. What do you think about having `['7.2', '7.4', '8.0']`?

LongTrong-exe: But phpunit 9 requires php 7.3 or higher

LongTrong-exe: @jmaas-onoffice I think we can skip run unittest at php7.2

jmaas-onoffice: Then I think you need to revert 6fc5136a46aba4b15426477f3017cacf0f81d497. We can just update the minimum PHP version and make a new major release.

.github/workflows/unit-tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest]
-        php-versions: ['7.2','7.3', '7.4']
+        php-versions: ['7.2', '7.4', '8.0']
```
jmaas-onoffice, line 19: I think it's better to revert 6fc5136a46aba4b15426477f3017cacf0f81d497. We will update the minimum PHP version.

jmaas-onoffice (approved):

jmaas-onoffice (dismissed):

jmaas-onoffice (approved):

jmaas-onoffice (dismissed):

jmaas-onoffice (approved):

PR #255, opened 2022-05-30
Test development .zip action

onOffice-Web:
In #253 I couldn't test the development .zip action, because the build always uses the workflow files from `master`.


jmaas-onoffice (approved):

jmaas-onoffice:
Works.

PR #109, opened 2021-08-20
P#60741 / P#52107 Update allow sorting by all field + P#55975 improve sorting controls

hungnc89:
fix #152

PR #258, opened 2022-05-31
P#60739 Remove unsupported data types - Data types still visible in some lists revert

yeneastgate:
…ve field


hungnc89 (approved):

LongTrong-exe (approved):

LongTrong-exe (approved):

hungnc89:
- Tester tested
- Video:

https://user-images.githubusercontent.com/72601287/174043037-46ae7d78-367e-4349-b7a5-d4043a40da5d.mp4

jmaas-onoffice (approved):

jmaas-onoffice (commented):
It seems that with v3.2 I already cannot select the unsupported data types in the field selection. This makes it difficult to test this change.

Do you have a version where the unsupported types can be added to the field list so that I can test the update?

jmaas-onoffice:
I'm a bit confused because as far as I can see this change has not been released. So how could it be that the issue is fixed, but we still have this PR open?

yeneastgate:
> I'm a bit confused because as far as I can see this change has not been released. So how could it be that the issue is fixed, but we still have this PR open?

The reason is that in the "loadLanguage()" function in the "Fieldnames.php" file, "'realDataTypes' => true".
This issue was fixed in pull request #109.
image

yeneastgate:
> It seems that with v3.2 I already cannot select the unsupported data types in the field selection. This makes it difficult to test this change.
> 
> Do you have a version where the unsupported types can be added to the field list so that I can test the update?

you can return master and comments line code on "FieldLoaderGeneric.php" and "Fieldnames.php"
image
image

jmaas-onoffice (approved):

jmaas-onoffice:
Ok, if I understand this PR correctly, it only adds some code so that on installation all those unsupported fields are removed, if they are used in a form, list or estate detail. Is that correct?

If so, I think maybe it's best if we close this PR, because testing those database deletions is not easy. At the same time, the fields are already not selectable for the fieldlist anymore. So the worst thing that can happen is that someone had added one of the unsupported fields a long time ago and it still is in their field list. But since the field is not output on the frontend, nothing breaks.

All in all, if I understand it correctly, I think it's best if we close this PR without merging, because the underlying problem has already been fixed. Sorry for not realizing this sooner!

PR #266, opened 2022-06-03
P#59851 New contact person fields not translated

LongTrong-exe:


LongTrong-exe:
https://user-images.githubusercontent.com/101325858/171787681-923f2876-7c7a-4807-ac9a-a2e3ccd57a0b.mp4

jmaas-onoffice (approved):
Wow, that was quick. Thank you!

jmaas-onoffice (changes requested):
Unfortunately, when I reload the page, some of the fields on the right show the English categories.

grafik

1. The bottom 7 fields have English categories when they should have German categories, except for "Alle E-Mail-Adressen" which is correct, because I removed and readded it. So when you add a field, it gets the correct German category, but when you reload, it jumps back to English.
2. Similarly, "Bild" has "(unbenannte Kategorie)" on the left, but when I reload, it gets "Formspezifische Felder" on the right.

Could you make sure that the categories stay consisten?

plugin/Field/Collection/FieldLoaderGeneric.php
```diff
@@ -75,6 +75,7 @@ public function load(): Generator
 	{
 		$newAddressFields = FieldModuleCollectionDecoratorReadAddress::getNewAddressFieldsWithTableNameKey();
 		$result           = $this->sendRequest();
+		$screen           = get_current_screen();
```
jmaas-onoffice, line 78: Why do you need to know the current screen here? I'm unsure what should depend on that.

jmaas-onoffice (changes requested):

plugin/Field/Collection/FieldLoaderGeneric.php
LongTrong-exe: This function is called from the config field of the form, estate and address. So I added "detail estate" screen check so that my new code doesn't affect other config fields causing unexpected errors.

jmaas-onoffice: Are you sure that this won't lead to issues later on? I find it quite unexpected that the field depends on the page, so this might be confusing later.

What is the worst that could happen if we didn't restrict it to the detail page?

LongTrong-exe: It doesn't seem to cause as many errors as I thought, I'll remove it. The code inside the `if screen` I used to move the fields in the `(unnamed category)` to the `Form Specific Fields`, but now it's giving me some errors in the form. I need more time to fix it

image

LongTrong-exe: But if you are in a hurry, I think the current code is ok.

jmaas-onoffice: Let's take the time to get this right. :)

If you find more opportunities to clean up and refactor the code, please do so! ;)

jmaas-onoffice: I was aiming to do a release next Tuesday, because that is when onOffice enterprise has a release, but there is no real reason why. So let's do the release when this is ready and take the time we need.

jmaas-onoffice (approved):
It's a bit difficult to understand the changes, because I'm not familiar with the code, but I'll test it.

jmaas-onoffice (changes requested):
I've still found a few inconsistencies:

1. In the interest form, there are two "Form-specific fields". (I've edited the HTML to have them both opened.)

    grafik
2. Is there a reason that "Message" is in its own category or should we put it into "Form Specific Fields"? The example image is from an interest form, but the field is present on all forms types except interest search forms.

    grafik
3. Why are the "Form Specific Fields" not at the top for owner forms like they are for the other form types?

    grafik
4. This is just something I realized: I would like to change the name "Form Specific Fields" to "Special Fields".

yeneastgate:
> Could you have a look at how to resolve these? :)

yes, I will check and research solution to resolve these

hungnc89 (approved):

hungnc89 (approved):

yeneastgate:
> I've still found a few inconsistencies:
> 
> 1. In the interest form, there are two "Form-specific fields". (I've edited the HTML to have them both opened.)
>    grafik
> 2. Is there a reason that "Message" is in its own category or should we put it into "Form Specific Fields"? The example image is from an interest form, but the field is present on all forms types except interest search forms.
>    grafik
> 3. Why are the "Form Specific Fields" not at the top for owner forms like they are for the other form types?
>    grafik
> 4. This is just something I realized: I would like to change the name "Form Specific Fields" to "Special Fields".
> 
> Could you have a look at how to resolve these? :)

I fixed all issues. pls, review the again code. This video test

Uploading New contact person fields not translated.mp4…

jmaas-onoffice (approved):

jmaas-onoffice:
Nice! I'm very happy with the improvements that we made. 👍

PR #269, opened 2022-06-07
P#72883 Uninstall does not clear all settings

dai-eastgate:


jmaas-onoffice:
Do you think there is a way to ensure that we don't forget to remove options on uninstall?

LongTrong-exe:
@jmaas-onoffice I think use sql query to get all options in db with prefix "onoffice" or "oo".

jmaas-onoffice:
That seems a bit dangerous. Is there a way to handle this in the code? For example a way to register new field that automatically remembers them for deletion when the plugin is uninstalled? Or a test that installs the plugin and uninstalls it and checks the database?

If this is too much work for too little value, that's ok. I am just curious. ;)

dai-eastgate:
Do you want to store all options that we registered as a json array? In the uninstall we can loop through that array & delete one by one and finally we delete the option that holds all options.

jmaas-onoffice:
I don't know what would be a good approach. 😅

Maybe it's best if we merge this, because this definitely works. And if you happen to find an improvement, we can always make a new PR. :)

PR #271, opened 2022-06-09
P#72891 "GDPR status" field cannot be renamed

dai-eastgate:


dai-eastgate:
Hi @jmaas-onoffice, while testing this issue, I have found 2 issues.
1. When I add the "Last Contact" field from "Contact" category, the ajax request will be throw server error and we can't see the saved alert. https://user-images.githubusercontent.com/106214469/172777145-e67244ba-b6ee-489a-a06a-7af6e73ba36f.mp4
2. When I add fields and then enter the custom label for those fields and finally click the save changes button. But when you refresh the page, the custom labels can't save. I can't save the custom label when you add the new field. Refresh the page and add the custom labels it will work as normal. 
https://user-images.githubusercontent.com/106214469/172778769-5d6f9b6a-31ba-4b2b-af01-2ce7f7640f4c.mp4

Do I need to fix those issues in this PR? Please give me some advices. Thank you.

dai-eastgate:
Hi @jmaas-onoffice, what is the type of the "Last Contact" field? Currently that type is "date", but I think the type here should be range type.

jmaas-onoffice:
I'm not sure I understand what you mean. Why should it be a range type?

dai-eastgate:
It has the from-to values. What type should we input to those fields?

jmaas-onoffice:
I'm still not sure. What problem is it that made you ask this question?

dai-eastgate:
With the date type we will save as a single value (number, text, ...), but the value of the "Last Contact" field is an array [min, max]. That why you see the argument error.

jmaas-onoffice:
@dai-eastgate Last Contact (`letzter_Kontakt`) should definitely be a date, from what I can tell.

Maybe there is a bug that makes the backend think its a range? Do you have an idea why it would show up as a range in the settings?

dai-eastgate:
image
@jmaas-onoffice on the frontend we setting the inputs name like this and when the server transforms data it will store as an array variable (I see some other fields have the same input name like [fieldName][min, max] and it will transform to an array variable).

jmaas-onoffice:
Linh informed me that the problem may be with what the API returns as a type for this. I will investigate this as soon as I find the time. 👍

jmaas-onoffice:
The API seems to return the correct type:

```
{
"label": "Adressen",
"letzter_Kontakt": {
  "type": "datetime",
  "length": null,
  "permittedvalues": null,
  "default": null,
  "compoundFields": [],
  "label": "Letzter Kontakt"
}
```

Can you investigate if the plugin is receiving the same?

dai-eastgate:
Yes, the plugin is receiving the same. There is the conflict between FE & the plugin, we are rendering two fields on the FE & it will send an array to the BE. BE code base is stored a single value for the datetime type.

jmaas-onoffice:
Ok, I'm a bit confused. Can we fix the issue? 😅

dai-eastgate:
What is the approach for this issue? We have two options:

1. Correct fields on FE. Do the FE render the fields correctly (one or two is correct)?
2. Change the code base on BE. we need separate two fields (different field name).

LongTrong-exe (approved):

js/onoffice-default-form-values.js
```diff
@@ -191,12 +191,15 @@ onOffice.default_values_input_converter = function () {
...
         if (!fieldDefinition.rangefield &&
             [
-                'integer', 'float', 'date',
+                'integer', 'float', 'date', 'datetime',
                 'urn:onoffice-de-ns:smart:2.5:dbAccess:dataType:float',
                 'urn:onoffice-de-ns:smart:2.5:dbAccess:dataType:int',
                 'urn:onoffice-de-ns:smart:2.5:dbAccess:dataType:integer',
                 'urn:onoffice-de-ns:smart:2.5:dbAccess:dataType:decimal'
             ].indexOf(fieldDefinition.type) >= 0) {
+            if (fieldDefinition.type === 'date' || fieldDefinition.type === 'datetime') {
```
jmaas-onoffice, line 200: Is this check redundant with the previous check?

plugin/Types/FieldsCollection.php
```diff
@@ -111,6 +111,15 @@ public function getFieldByModuleAndName(string $module, string $name): Field
...
 	public function merge(FieldModuleCollection $pFieldsCollection, string $fallbackCategoryName = '')
 	{
+		$changeDefaultFields = [
+			'DSGVOStatus'          => 'GDPR status',
+			'SpracheKontakt'       => 'Language',
+			'KontaktformBevorzugt' => 'Preferred form of contact',
+			'gwgVertragspartner'   => 'Contractual partner',
+			'gwgAusweisart'        => 'ID type',
+			'newsletter_aktiv'     => 'Newsletter',
+			'workContract'         => 'Employment',
```
jmaas-onoffice, line 121: Does this need to be translated?

templates.dist/fields.php
```diff
@@ -133,7 +134,7 @@ function renderFormField(string $fieldName, onOffice\WPlugin\Form $pForm, bool $
 		if (\onOffice\WPlugin\Types\FieldTypes::FIELD_TYPE_SINGLESELECT == $typeCurrentInput) {
 			$output .= '<select class="custom-single-select" size="1" name="' . esc_html($fieldName) . '" ' . $requiredAttribute . '>';
 			/* translators: %s will be replaced with the translated field name. */
-			$output .= '<option value="">' . esc_html(sprintf(__('Choose %s', 'onoffice'), $fieldLabel)) . '</option>';
+			$output .= '<option value="">' . esc_html(sprintf(__('Choose %s', 'onoffice'), $fieldDefault)) . '</option>';
```
dai-eastgate, line 137: This line already covered the translation

jmaas-onoffice: So `$fieldDefault` is translated? Then it's all good. :)

js/onoffice-default-form-values.js
jmaas-onoffice: I just though that

```php
[/*...*/ 'date', 'datetime' /*...*/].indexOf(fieldDefinition.type) >= 0
```

is the same as

```php
fieldDefinition.type === 'date' || fieldDefinition.type === 'datetime'
```
But I leave it up to you. :)

templates.dist/fields.php
dai-eastgate: Yes

js/onoffice-default-form-values.js
dai-eastgate: If you use the previous check, the input text will be apply for the integer type also. That why we need another condition to render the correct input type.

templates.dist/fields.php
jmaas-onoffice: Yes! And I think the problem here is that the translations for the labels should not come from WordPress but from onOffice enterprise.

I think #266 was a similar situation, so maybe you can find information there.

js/onoffice-default-form-values.js
jmaas-onoffice: Oh, sorry! Now I see it. For some reason I didn't see that the inner if only covers one line and that there are more after it! 🤦

templates.dist/fields.php
dai-eastgate: API doesn't return the default field. that why I need to create a default array for the name of option default.

dai-eastgate: Do you have any update on this issue.

jmaas-onoffice: Thanks for the reminder, I forgot to answer!

What is this default then? What do you use it for?

dai-eastgate: The default is the un-change label likes "GDPR status". API doesn't return the original label (default), so I make a hard code default fields to list all original labels.

jmaas-onoffice: I think I'm not understanding what you mean. I can change the label in onOffice enterprise and it will change on the frontend as well.

grafik

Let's clear this up in the weekly on Wednesday. ;)

dai-eastgate: This is the bug you are mention. We should keep the default label for the default option of a select box. But in the API response I can't find some information about "GDPR status", that why I need to make a hardcode list (for the default option name).

dai-eastgate: image
This is your expectation. Please correct me if I'm wrong. Do we need to keep the default option of the select as the original label or it is just a mistake from your expectation so we need apply the change for the default option also?

jmaas-onoffice: As discussed in the weekly, this is a mistake in my requirements. Both the label and the dropdown should show the same label.

dai-eastgate:
@jmaas-onoffice there is no change to fix this bug. All the bugs already resolved in other issues. Could you please re-test it and then we can close this PR?

jmaas-onoffice:
You're right, I just tested it. Well, this is also a nice way to solve a bug. :D

Thanks for noticing and informing me! :)

PR #279, opened 2022-06-20
P#63827 Sent from override email instead of default email

LongTrong-exe:


jmaas-onoffice (changes requested):
It feels like there is some logic duplication. Is there a way to refactor this?

plugin/FormPost.php
```diff
@@ -136,6 +136,19 @@ public function initialCheck(DataFormConfiguration $pConfig, int $formNo)
 		}
 	}
 
+	/**
+	 * @param  DataFormConfiguration  $pFormConfiguration
+	 *
+	 * @return bool|mixed|void
+	 */
+	protected function getRecipient( DataFormConfiguration $pFormConfiguration ) {
```
jmaas-onoffice, line 144: Why is the method attached to this class? It only uses the `DataFormConfiguration`, right?

jmaas-onoffice (changes requested):

plugin/FormPost.php
LongTrong-exe: Right, what do you mean? If you don't want to write there, I think I can write this function in `DataFormConfiguration` class instead of in `FormPost`.

jmaas-onoffice (approved):

jmaas-onoffice:
Nice, looks like you managed to reduce a lot of code with the same logic. 👍

PR #288, opened 2022-06-30
P#71175 Simplify translations

dai-eastgate:


LongTrong-exe (dismissed):

jmaas-onoffice (changes requested):
I noticed one instance where the translation function is used with a variable:

https://github.com/onOffice-Web-Org/oo-wp-plugin/blob/d793ae774e68f7274cc372dc0c38748387cfdd47/plugin/Gui/DateTimeFormatter.php#L96

This is problematic, because the translation software will not be able to detect this. See http://ottopress.com/2012/internationalization-youre-probably-doing-it-wrong/ for more information.

dai-eastgate:
I just removed the translation for variable `$format`. We can't translate the variable, so just leave it there.

jmaas-onoffice (approved):

jmaas-onoffice (approved):

jmaas-onoffice:
With this change, we do not need to keep updating the files in https://github.com/onOffice-Web-Org/oo-wp-plugin/tree/bdb6c937c42606e24cb58851c2b6b5009a15f72b/languages. All translations will be done via the WordPress translation platform.

Could we deactivate all scripts that update these language files? The files need to be kept for old installations to still work, but we do not need to update them anymore.

jmaas-onoffice (changes requested):
Looks good!

I would only suggest one more thing: https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/288#issuecomment-1196451240

dai-eastgate:
We have script "make pot" to generate the pot file but we need to run it manually. So we just leave it there from now, don't need to run it when we use the translation anymore.

jmaas-onoffice:
Nice! I removed the `pot`-target from the Makefile. Is that alright?

jmaas-onoffice (approved):

jmaas-onoffice:
@dai-eastgate Can you double-check that removing the `pot`-file is ok? Just for my sanity. :)

dai-eastgate:
I checked it. It won't generate a pot file anymore (keep the old one).

PR #315, opened 2022-08-23
P#77244 Template groups are not translated

dai-eastgate:


LongTrong-exe (approved):

LongTrong-exe (approved):

jmaas-onoffice (changes requested):
Since we merged #288, the language files (.mo and .po) should not be updated anymore.

jmaas-onoffice (approved):

PR #317, opened 2022-08-25
P#77246 Wrong use of translation function

dai-eastgate:
related to issue https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/306

chaged logs:


LongTrong-exe (approved):

LongTrong-exe (approved):

plugin/Field/FieldModuleCollectionDecoratorInternalAnnotations.php
```diff
@@ -64,7 +64,7 @@ public function getFieldByModuleAndName(string $module, string $name): Field
 		$pField = parent::getFieldByModuleAndName($module, $name);
 		$newLabel = $this->getNewFields()[$module][$name] ?? null;
 		if ($newLabel !== null) {
-			$pField->setLabel(__($newLabel, 'onoffice-for-wp-websites'));
+			$pField->setLabel( sprintf( __( '%s', 'onoffice-for-wp-websites' ), $newLabel ) );
```
jmaas-onoffice, line 67: How is this supposed to work? How should the WordPress translation know what `$newLabel` is to enable translation? :)

jmaas-onoffice (changes requested):

plugin/Field/FieldModuleCollectionDecoratorInternalAnnotations.php
dai-eastgate: it has been translated, for example line 49 just $newLabel is enough, it needs to be translated before using it.
Can you please test it? I tested it and it still translates normally :)
As for the words that have not been translated before, I will check them carefully in this PR

jmaas-onoffice: If `$newLabel` is already translated, why are we using `__( '%s', 'onoffice-for-wp-websites' )` here? :)

jmaas-onoffice (approved):

dai-eastgate:
@jmaas-onoffice Sorry, I forgot to submit a review request here

PR #318, opened 2022-08-25
P#77242 Missing translations in interest form template

dai-eastgate:
related to issue https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/304

chaged logs:

### DIFF
--- a/templates.dist/form/applicantform.php
+++ b/templates.dist/form/applicantform.php
@@ -74,7 +74,7 @@ if ($pForm->getFormStatus() !== \onOffice\WPlugin\FormPost::MESSAGE_SUCCESS) {
 		</div>
 	</p>
 	<p>
-	<h1>Ihre Suchkriterien</h1>
+	<h1><?php esc_html_e('Your search criteria', 'onoffice-for-wp-websites'); ?></h1>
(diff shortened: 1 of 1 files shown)


LongTrong-exe (approved):

LongTrong-exe (approved):

templates.dist/form/applicantform.php
```diff
@@ -68,7 +68,7 @@
 
 ?>
 	<p>
-	<h1>Ihre Kontaktdaten</h1>
+	<h1><?php esc_html_e('Your contact details', 'onoffice'); ?></h1>
```
jmaas-onoffice, line 71: This is the wrong text domain. See #288.

jmaas-onoffice (changes requested):

templates.dist/form/applicantform.php
dai-eastgate: @jmaas-onoffice I fixed it. Please review again

jmaas-onoffice (changes requested):
There are still parts of the template that are not translated.

jmaas-onoffice (approved):

PR #324, opened 2022-09-06
P#77250 Offer all estate status fields in the template

dai-eastgate:
related to issue https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/308

chaged logs:
-

### DIFF
--- a/templates.dist/estate/default.php
+++ b/templates.dist/estate/default.php
@@ -27,7 +27,22 @@ use onOffice\WPlugin\Favorites;
 require 'SearchForm.php';
 /* @var $pEstates onOffice\WPlugin\EstateList */
 
-$dontEcho = array("objekttitel", "objektbeschreibung", "lage", "ausstatt_beschr", "sonstige_angaben");
+$dontEcho = array(
+	"objekttitel",
+	"objektbeschreibung",
+	"lage",
+	"ausstatt_beschr",
+	"sonstige_angaben",
+	"referenz",
+	"reserviert",
+	"verkauft",
+	"exclusive",
+	"neu",
+	"top_angebot",
+	"preisreduktion",
+	"courtage_frei",
+	"objekt_des_tages"
+);
 
 ?>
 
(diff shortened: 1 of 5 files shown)


LongTrong-exe (approved):

templates.dist/estate/default.php
```diff
@@ -61,6 +61,15 @@
 	while ( $currentEstate = $pEstatesClone->estateIterator() ) :
 		$marketingStatus = $currentEstate['vermarktungsstatus'];
 		unset($currentEstate['vermarktungsstatus']);
+		unset($currentEstate['referenz']);
```
jmaas-onoffice, line 64: I'm curious: Why are we unsetting the values here? Are they conflicting with something?

dai-eastgate: They don't conflict. It's not in use at the moment so I just hid it
Do you want to display it or use it in any way?
image

jmaas-onoffice: Ah, ok! I forgot that they get rendered.

There is also the option of adding those fields to the [`dontEcho` array](https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/324/files/79ce57d0bf507643e66912f02d29e1207a44ca16#diff-fe5dbe760501e4e626f9343c784248ce823bead31e4209da309b956045454e9cR30). ;)

dai-eastgate: I added it you can review. :)

jmaas-onoffice (approved):

jmaas-onoffice (changes requested):
image

jmaas-onoffice (approved):

jmaas-onoffice:
I made a mistake with the requirements: This is backwards incompatible.

When users update, they would need to change their personalized templates so that the individual fields do not get output. We need to find a mechanism so that no template changes are needed on update.

Since this is quite a difficult thing to do properly, I will close this PR for now. Sorry for that!

PR #439, opened 2023-02-03
Unescape the POST fields

jmaas-onoffice:
This fixes an issue where saving custom labels in forms or estates with quotes would add backslashes. Saving again would add even more backslashes.


jmaas-onoffice:
@dai-eastgate Could someone review this PR? :)

PR #440, opened 2023-02-03
Fix InvalidPathException in child theme

jmaas-onoffice:
When a template from the parent theme folder is selected and you switched to a chlid theme, the code would throw an exception because the template was not found in the child theme.

This change lets the plugin find the template in the parent folder.

Note that when using a child theme, the parent theme's templates will not be displayed in the backend. This quick fix has the goal to prevent an unnecessary crash. Proper child theme support needs to be implemented separately.


jmaas-onoffice:
@dai-eastgate Could someone review this PR? :)

dai-eastgate:
> @dai-eastgate Could someone review this PR? :)

Yes, I am checking it

PR #456, opened 2023-02-21
30460 Make detail view shortcode work in ACF blocks

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/448

changed log :
replase shortcode detail


jmaas-onoffice:
See https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/448#issuecomment-1439962186.

PR #459, opened 2023-03-06
31368 Property search button value displays all properties

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/discussions/452

changed log :
update property search reference estate


andernath (commented):
Codereview and test ✔

PR #471, opened 2023-03-21
32003 Honeypot backwards compatibality

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/discussions/463

changed log :
update honeypot backwards compatibality


dai-eastgate:
@fredericalpers 
We have found solution to resolve "Backwards compatibiliy is not given when the honeypot option is activated by default on existing installations."
Although the honeypot option is activated, it still does not affect If the client has custom CSS the field is visible in frontend.
So we don't need to deactivate it.
please watch the video and give me your opinion on the matter?

https://user-images.githubusercontent.com/106214469/226825062-079ff7c5-5a46-48b3-89fd-d07a1984ebd2.mp4

PR #509, opened 2023-04-27
32989 Clearing the onOffice cache

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/504

changed log :
Notice when clearing cache

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -339,7 +339,7 @@ add_action('parse_request', function () use ( $pDI ) {
 add_action('admin_notices', function () {
 	if (get_option('onoffice-notice-cache-was-cleared') == true) {
 		$class = 'notice notice-success is-dismissible';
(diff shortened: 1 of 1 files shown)


plugin.php
```diff
@@ -330,11 +330,21 @@ function update_status_close_action_button_option()
...
 		wp_safe_redirect($location);
 		exit();
 	}
 });
 
+add_action('admin_notices', function () {
+	if (get_option('onoffice-notice-cache-was-cleared') == true) {
+		$class = 'notice notice-success is-dismissible';
+		$message = __('The cache was cleared successfully.', 'onoffice-for-wp-websites');
+		printf('<div class="%1$s"><p>%2$s</p></div>', esc_attr($class), $message);
```
andernath, line 343: Wouldn't be using esc_html for $message be more safe?

andernath (approved):

PR #519, opened 2023-05-09
34277 Custom label of the "message" field doesn't work

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/514

changed log :
update custom label


andernath (approved):
I will start testing

Issue #516, opened 2023-05-03
Map in property listing


onOffice-Web:
### Discussed in https://github.com/onOffice-Web-Org/oo-wp-plugin/discussions/505

<div type='discussions-op-text'>

<sup>Originally posted by **fredericalpers** April 25, 2023</sup>
### Current state

As soon as an address is shared to the website, the map will appear above the property list. Currently it's only possible to deactivate the map above the property list with a template customization.

### Desired state

The customers should be able to toggle the map “on” or “off” with a checkbox in the WP Backend. (See attached screenshot "listing-layout-map") It should be possible to hide the map, although the address is released, so that the map is only shown in the detail page of a property. The checkbox should be activated by default.

### Investigate

In addition to the above mentioned changes please investigate the following:

Show all properties on the map and not only those that are currently loaded on page 1 or when on page 2 only the properties that are currently loaded on page 2. (See attached screenshot "map-pins")

![listing-layout-map](https://user-images.githubusercontent.com/88546396/234267086-48b833b0-fb81-4bdf-a4bd-56358fcf4e8d.jpg)

![map-pins](https://user-images.githubusercontent.com/88546396/234266193-34479ee7-18f3-4408-8556-8d8c417577f8.PNG)
</div>

yeneastgate:
@fredericalpers I added a checkbox to on/off the map in the “Estate List”. 
On the map, all of the properties that have published addresses have been shown the same for all pagination.
Please watch the attached video: 

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/1ec9890f-36e5-452f-a0bc-6be526021047

Also, I would like to confirm that if the customer has a large number of Estates, then these Estates will be shown all on the map? 
Please also let me know your opinion about that. Thanks.

fredericalpers:
> @fredericalpers I added a checkbox to on/off the map in the “Estate List”. On the map, all of the properties that have published addresses have been shown the same for all pagination. Please watch the attached video:
> 
>  map_in_property_listing.mp4 
> Also, I would like to confirm that if the customer has a large number of Estates, then these Estates will be shown on the map? Please also let me know your opinion about that? Thanks.

thank you! 

to confirm: yes, all properties, where the customer is sharing the adress, should be visible on the map.

yeneastgate:
[...]

PR #523, opened 2023-05-18
34563 Map in property listing

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/516

changed log :
Add checkbox to show map in estate list


yeneastgate:
> @yeneastgate please see my comment in #516

Yes, I got it. Thanks!

PR #524, opened 2023-05-19
34775 Rename "Reduce values by selected filter & filterable"

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/517

changed log :
Rename reduce values by selected filter and filterable


andernath (approved):
I will start testing

PR #531, opened 2023-05-26
34982 Edit list view

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/529

changed log :
remove screen option is not used


andernath (approved):
I will start testing

PR #532, opened 2023-05-26
34981 Saving lists or forms without a name

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/527

changed log :
add notice when saving lists or forms without a name.

### DIFF
--- a/plugin/Gui/AdminPageAddressListSettings.php
+++ b/plugin/Gui/AdminPageAddressListSettings.php
@@ -380,7 +380,7 @@ class AdminPageAddressListSettings
 	protected function checkFixedValues($row)
 	{
 		$table = RecordManager::TABLENAME_LIST_VIEW_ADDRESS;
-		$result = isset($row[$table]['name']) && $row[$table]['name'] != null && !empty(trim($row[$table]['name']));
+		$result = isset($row[$table]['name']) && !empty(trim($row[$table]['name']));
 
 		return $result;
 	}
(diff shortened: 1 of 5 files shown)


plugin/Gui/AdminPageAddressListSettings.php
```diff
@@ -366,7 +380,7 @@ public function doExtraEnqueues()
 	protected function checkFixedValues($row)
 	{
 		$table = RecordManager::TABLENAME_LIST_VIEW_ADDRESS;
-		$result = isset($row[$table]['name']) && $row[$table]['name'] != null;
+		$result = isset($row[$table]['name']) && $row[$table]['name'] != null && !empty(trim($row[$table]['name']));
```
andernath, line 383: $row[$table]['name'] != null 
Is this necessary or is not null already checked with isset?

Same question for other classes below.

andernath (commented):
Finished review and have one question. Please check. @yeneastgate :)

plugin/Gui/AdminPageAddressListSettings.php
yeneastgate: @andernath I will check and let you know soon.

yeneastgate: @andernath In this case, the isset function already checks undefined and null => so $row[$table]['name'] != null is not needed.
I will check and delete it

andernath (approved):

PR #560, opened 2023-06-14
35071 Ownerform - Create address and property

dai-eastgate:


andernath (approved):
Review ✔
I will start testing.

PR #572, opened 2023-06-19
35110 "Geo range search" - the coordinates in the property tiles can not be hidden

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/539

changed log :
Update geo range search


andernath (approved):
I will start testing

PR #586, opened 2023-07-03
36779 Sort property lists

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/549

changed log :
Add sort property lists


andernath (approved):
I will start testing

PR #592, opened 2023-07-11
36781 onoffice-style.css with child theme

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/548

changed log :
update onoffice-style.css with child theme


andernath (approved):
I will start testing.

PR #594, opened 2023-07-17
36784 Plugin can not set supervisor in address record

dai-eastgate:


plugin/Form/FormAddressCreator.php
```diff
@@ -68,7 +68,7 @@ public function __construct(
 	 * @throws NotFoundException
 	 */
 	public function createOrCompleteAddress(
-		FormData $pFormData, bool $mergeExisting = false, string $contactType = ''): int
+		FormData $pFormData, bool $mergeExisting = false, string $contactType = '', int $estateId = null): int
```
andernath, line 71: Could you please add PHPDoc for these arguments: $contactType and $estateId for complete documentation? :)

plugin/Form/FormAddressCreator.php
```diff
@@ -82,6 +82,13 @@ public function createOrCompleteAddress(
 		if ( key_exists( 'newsletter', $requestParams ) ) {
 			unset( $requestParams['newsletter'] );
 		}
+		if (!empty($estateId)) {
+			$userName = $this->getSupervisorUsernameByEstateId($estateId);
+			if (!empty($userName)) {
```
andernath, line 87: In the past we got minor bugs with using empty() function because "0" and 0 is also empty true. It is unlikely but what happens if the ID or username is 0?

Do you think using isset could be better or could causes other errors?

Edit:
Oh I see getSupervisorUsernameByEstateId returns maybe an empty string. So I guess we need empty() here.

andernath (commented):
@dai-eastgate 
Looks good. I just found a missing part in documentation. Please take a look. :)

plugin/Form/FormAddressCreator.php
dai-eastgate: I'll check and fixed it asap

dai-eastgate:
@andernath I fixed it, please review again for me. Thanks

andernath (approved):
Thanks, I will start testing.

PR #595, opened 2023-07-18
36782 Prevent double form submission

yeneastgate:


andernath (commented):
@yeneastgate 
Please take a look in the review. Thank you!

yeneastgate:
> @yeneastgate Please take a look in the review. Thank you!

yes, I will check it

yeneastgate:
@andernath After implementing the 'add class' step for the selector, I had a question about whether you want me to "add class" in a template like the attached image or use JavaScript to "add class".
image

andernath:
> @andernath After implementing the 'add class' step for the selector, I had a question about whether you want me to "add class" in a template like the attached image or use JavaScript to "add class". 

@yeneastgate 
The place where you add these classes is correct. 
But I would like to change the class name, because it is very non-specific and could already be used. I will discuss in our team for a better naming.

andernath:
@yeneastgate
Please add .oo-form in all templates. And replace .oo-form-owner, .oo-form-applicant ... etc for your current class names.

I guess in this case you can use just .oo-form in honeypot js and double submit js as a selector. 
And maybe in future releases we could need a these more specific class name (like .oo-form-owner), thats why you can add them as well in the templates.

yeneastgate:
> @yeneastgate Please add .oo-form in all templates. And please replace the current class names with .oo-form-owner, .oo-form-applicant ... etc.
> 
> I guess in this case you can use just .oo-form in honeypot js and double submit js as a selector. And maybe in future releases we could need a these more specific class name (like .oo-form-owner), thats why you can add them as well in the templates.

yes, I will implement it.

yeneastgate:
@andernath I refactored the code for the honeypot file and prevented submission using the ".oo-form" class selector. Please watch the attached video and let me know your opinion about this. Thanks!
https://files.fm/u/ht43bjzj2

js/onoffice-honeypot.js
```diff
@@ -1,23 +1,41 @@
...
+
+    addHoneypotToForms(formElements);
+
+    function getFormsBySpecialHtmlAttribute() {
+        const formElements = [];
+        const forms = document.querySelectorAll('.oo-form, #onoffice-form');
+
+        forms.forEach(form => {
+            const dataAttr = form.getAttribute('data-applicant-form-id');
+            if (dataAttr === null) {
```
andernath, line 12: I think you are checking here if the form is an ApplicantSearchForm. 
Previously it was checked with "form.type !== 'applicantsearch'". I find that more readable and more stable than the new solution. Can we continue to use checking with form.type or did it not work?

Overall I like the refactoring very much. Good work. :)

js/onoffice-prevent-double-form-submission.js
```diff
@@ -0,0 +1,33 @@
...
+
+    handleFormSubmissions(formElements);
+
+    function getFormsBySpecialHtmlAttribute() {
+        const formElements = [];
+
+        const forms = document.querySelectorAll('.oo-form, #onoffice-form');
+        formElements.push(...forms);
+
+        return formElements;
```
andernath, line 12: Is the whole function getFormsBySpecialHtmlAttribute really necessary?
Wouldn't it be the same just to use const formElements = document.querySelectorAll('.oo-form, #onoffice-form'); ?

andernath (commented):
@yeneastgate Your solution looks good to me!
Just got 2 questions here for improvement, could you please take a look?

js/onoffice-honeypot.js
yeneastgate: @andernath If there are multiple forms on the same page, using form.type to check for an applicant search form will cause the honeypot feature to not work on other forms (contact, interest, owner).

js/onoffice-prevent-double-form-submission.js
yeneastgate: Yes, you're right. I will fix it.

yeneastgate:
Hi @fredericalpers @andernath 
- While implementing this issue, I found a bug: when I enable reCAPTCHA on a page that has multiple forms, I meet an error when trying to submit the second form or any subsequent ones. This issue exists in the master branch.
- The cause is when submitting the second form and beyond, the data does not receive g-recaptcha-response, which leads to reCAPTCHA identifying it as a bot. 
=> Cannot create an address and send an email
Video: 

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/3774adc4-7227-4734-b220-529cbc5b717d

js/onoffice-honeypot.js
andernath: @yeneastgate what a pity. :/ Thank you for info.

andernath:
> 
@yeneastgate 
Thank you for your thoughtful testing 👍. We have already recorded a task for this bug and will add it to a later release.

js/onoffice-prevent-double-form-submission.js
yeneastgate: I fixed it. Please review again for me. Thanks!

andernath (approved):
I will start testing

PR #618, opened 2023-08-22
39253 SEO Title and description

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/577

changed log :
Set limit character for title and description

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -66,6 +66,8 @@ use onOffice\WPlugin\Utility\__String;
 use onOffice\WPlugin\Utility\Redirector;
 use onOffice\WPlugin\WP\WPQueryWrapper;
 
+const DEFAULT_LIMIT_CHARACTER_TITLE = 60;
+
 define('ONOFFICE_DI_CONFIG_PATH', implode(DIRECTORY_SEPARATOR, [ONOFFICE_PLUGIN_DIR, 'config', 'di-config.php']));
 
 $pDIBuilder = new ContainerBuilder();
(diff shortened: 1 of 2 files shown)


plugin.php
```diff
@@ -399,4 +403,24 @@ function update_status_close_action_button_option()
 	};
 }, 500);
 
+function limit_characters(string $text, int $max_length): string
```
andernath, line 406: Why is this function necessary? Wouldn't _substr($result['title'], 0, DEFAULT_LIMIT_CHARACTER_TITLE);_ do the same?

andernath (commented):
@dai-eastgate I have a question in review. Please take a look.

plugin.php
andernath: @dai-eastgate 
Oh I guess you do not want to cut a word in half? Like 15 chars for "Beautiful cottage" would be "Beautiful" and not "Beautiful cott". Right?

I will clarify if this is necessary with the team.

andernath: The Seo Team thinks it's a good idea not to cut off words at the end.

I tried to improve the Code a bit. What do you think @dai-eastgate ? (sorry for bad formatting) 

`$result = $pDI->get(EstateViewDocumentTitleBuilder::class)->buildDocumentTitle($title);`
`if (strlen($result['title']) > DEFAULT_LIMIT_CHARACTER_TITLE) {`
      `$shortenedTitle = substr($result['title'], 0, DEFAULT_LIMIT_CHARACTER_TITLE);`
  `if (substr($result['title'], DEFAULT_LIMIT_CHARACTER_TITLE, 1) != ' ') {`
      `$shortenedTitle = substr($shortenedTitle, 0, strrpos($shortenedTitle, ' '));`
   `}`
   `$result['title'] = $shortenedTitle;`
`}`
`return $result;`

yeneastgate: @andernath yes,I will try the way you suggested, and will report back the result, thanks

dai-eastgate: @andernath yes, we will try the way you suggested, and will report back the result, thanks

dai-eastgate: @andernath I tried your suggestion, and it worked well. Please review the code I have refactored. Thank you!

andernath (approved):
I will start testing

PR #626, opened 2023-09-08
39622 Plugin uses deprecated API

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/607

changed log :
Api estate types replace "XMLHttpRequest" with "Fetch"


andernath (approved):
I will test

Issue #651, opened 2023-10-06
onoffice-prevent-double-form-submission.js


fredericalpers:
# Current state
The [onoffice-prevent-double-form-submission.js](https://github.com/onOffice-Web-Org/oo-wp-plugin/blob/master/js/onoffice-prevent-double-form-submission.js) script queries the class `.submit_button` and is then supposed to prevent the form from being submitted multiple times by setting the button to disabled.

However the [submit button](https://github.com/onOffice-Web-Org/oo-wp-plugin/blob/master/templates.dist/form/formsubmit.php#L53) (if reCAPTCHA is not used) does not have the class `.submit_button` and is therefore not disabled. If you add the class to the button, the form cannot be submitted. The form is not submitted prior to the button being disabled.

# Desired state

The javascript needs to be revised and improved so that submitting continues to be functional while multiple submissions are prevented

fredericalpers:
@dai-eastgate @yeneastgate please try to fix this during ramp up :)

yeneastgate:
@fredericalpers I would like to confirm some points with you:


> However the [submit button](https://github.com/onOffice-Web-Org/oo-wp-plugin/blob/master/templates.dist/form/formsubmit.php#L53) (if reCAPTCHA is not used) does not have the class `.submit_button` and is therefore not disabled. If you add the class to the button, the form cannot be submitted. The form is not submitted prior to the button being disabled.

**Case 1:** When not using reCAPTCHA
I don't use script queries in the ".submit_button" class in the "onoffice-prevent-double-form-submission.js" to prevent the form from being submitted multiple times by setting the button to disabled.
=> I use this logic

```
$(formElement).on('submit', function (event) {
        if (submitted) {
            event.preventDefault();
        } else {
            submitted = true;
        }
    });
```

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/a5f9489f-cef4-41d5-bcf6-27b4d461bdf7


=> You don't need to add the ".submit_button" class to the "submit button".
But we can still prevent the form from being submitted multiple times.
=> This issue #651 does not exist in the master branch.

Do you want to use script queries in the ".submit_button" class in the "onoffice-prevent-double-form-submission.js" to prevent the form from being submitted multiple when not using reCAPTCHA?

yeneastgate:
@fredericalpers 
**Case 2:** Using reCAPTCHA
[...]

PR #658, opened 2023-10-18
41417 do not disable "submit button" without reCAPTCHA

yeneastgate:
related to #651 
change log:
Update disable button submit without reCAPTCHA


andernath (approved):
I will start testing

PR #659, opened 2023-10-18
41389 Hotfix cannot submit with reCAPTCHA

yeneastgate:
related to #651 
change log:
Update submit form with reCAPTCHA


andernath (approved):
I will start testing

PR #665, opened 2023-10-31
41620 Implement City-Select into WP-Plugin

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/655
change log:
Add an option to convert from input type to select for the "city" field


andernath (dismissed):

fredericalpers:
@dai-eastgate unfortunately we still need to make some improvements here.

If I update the plugin with this feature, I can still use the checkbox "Hide empty values from onOffice enterprise" even though its supposed to be disabled. Currently it is deselected by default.
The styling is only corrected after a page reload with the browser console open.

If I do not reload the page with the browser console open and tick the checkbox "Hide empty values from onOffice enterprise" I get an error: The API returned an error with code 171 and the following message: No singleselect- or multiselect field.

To fix those issues we would like to adjust things as followed:

- [ ] Remove the "Hide empty values from onOffice enterprise" checkbox.
- [ ] All empty values should be removed by default.

The styling and handling differs from the other select fields. The location select is smaller and you can only select one value. See screenshot below.

- [ ] Adjust the styling to match the other select fields.

The order of the locations in the open Select seems arbitrary. An alphabetical order would probably be most helpful for the user.

- [ ] Add logic for alphabetically ordered locations.

Even though we initially wanted it to be a single select, it might make more sense to use a multiselect to be able to filter for more than one location at the same time.

- [ ] Change to multiselect, to be able to filter for more location at a time.


stylingissue

dai-eastgate:
> If I update the plugin with this feature, I can still use the checkbox "Hide empty values from onOffice enterprise" even though its supposed to be disabled. Currently it is deselected by default.
> The styling is only corrected after a page reload with the browser console open.

@fredericalpers yes, I will check it

dai-eastgate:
@fredericalpers I have completed 80% of the checklist, and I checked the parts I have completed. Please watch the demo video: https://files.fm/u/ecy2f3p5qn
Currently, I still have the checklist: 
> * Adjust the styling to match the other select fields.

I need 1 more day to check it
Because current input [type = "select"] is being applied "default styling" of select2 library

dai-eastgate:
> * Adjust the styling to match the other select fields.

@fredericalpers The styling of "The location select"(city field) is smaller because it is applied "default style" of the jQuery Select2 library
However, the styling of "The location select"(city field) is not smaller than other inputs (input[type=text], input[type=number]...), because these other inputs are being applied style of "active theme" in the WordPress backend.
=> So I think your requirement "Adjust styling to suit other selected fields" is not necessary.

I recommend that:
- Your customer should add styling for input has "type = select2" manually according to styling for the "input" HTML element of "the active theme" in the WordPress backend.
- ex: "The active theme" in the WordPress backend is "Royal Elementor Kit".
We add styling for input has "type = select2" manually according to styling for the "input" HTML element of Royal Elementor Kit
=> As you can see in the video, we can "adjust the styling to suit other selected fields": https://files.fm/u/4vbgtzeav9#/view/he98dxubyz

If we add styling for input that has "type = select2" manually on the onOffice WordPress plugin side => That will result in CSS attribute overwriting and conflict on the theme and onOffice WordPress plugin. Please check and let me know your opinion. Thanks!
image

andernath:
> > * Adjust the styling to match the other select fields.
> 
> @fredericalpers The styling of "The location select"(city field) is smaller because it is applied "default style" of the jQuery Select2 library However, the styling of "The location select"(city field) is not smaller than other inputs (input[type=text], input[type=number]...), because these other inputs are being applied style of "active theme" in the WordPress backend. => So I think your requirement "Adjust styling to suit other selected fields" is not necessary.
> 
> I recommend that:
> 
> * Your customer should add styling for input has "type = select2" manually according to styling for the "input" HTML element of "the active theme" in the WordPress backend.
> 
> image
> 
> * ex: "The active theme" in the WordPress backend is "Royal Elementor Kit".
>   We add styling for input has "type = select2" manually according to styling for the "input" HTML element of Royal Elementor Kit
>   => As you can see in the video, we can "adjust the styling to suit other selected fields": https://files.fm/u/4vbgtzeav9#/view/he98dxubyz
> 
> If we add styling for input that has "type = select2" manually on the onOffice WordPress plugin side => That will result in CSS attribute overwriting and conflict on the theme and onOffice WordPress plugin (look at the example image below). Please check and let me know your opinion. Thanks! image

@dai-eastgate Thank you for your research.
I agree with you, style needs to be adjustet in customer Theme. 
I will start review and testing.

andernath (approved):

PR #666, opened 2023-11-02
41664 floating save button

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/635
change log:
add a floating save button


css/admin.css
```diff
@@ -496,4 +496,66 @@ body #post-body-content p.wp-clearfix:nth-child(1){
 		font-size: 14px;
 		white-space: nowrap;
 	}
-}
\ No newline at end of file
+}
+
+@media only screen and (max-width: 1365px) {
+	#block-publish .postbox-header {
```
andernath, line 502: Please use classes for styling if possible

andernath (dismissed):

andernath:
Thanks for your changes @yeneastgate.
While updating Plugin Version 4.17 to onoffice-4.17-10-g6142b5e1 for testing I had some cache problems with the new styling. 
I think we have to increase the CSS version in AdminViewController.php line 316 from 4.2.0 to 4.18 or?

yeneastgate:
> Thanks for your changes @yeneastgate. While updating Plugin Version 4.17 to onoffice-4.17-10-g6142b5e1 for testing I had some cache problems with the new styling. I think we have to increase the CSS version in AdminViewController.php line 316 from 4.2.0 to 4.18.0 or?

@andernath I think we should leave it at the latest version, do you agree?
Can you review it?

andernath (approved):

PR #671, opened 2023-11-07
41980 Form can not be sent with unfilled required fields and enabled recaptcha

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/669
changed log:
Change logic multiple form recaptcha


andernath (approved):
I will start testing

PR #680, opened 2023-11-20
42302 Can't remove/delete secrets and keys

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/677
changed log:
Add delete button and show keys


js/onoffice-handle-notification-actions.js
```diff
@@ -15,4 +15,19 @@ jQuery(document).ready(function ($) {
 
 		jQuery.post(warning_active_plugin_vars.ajaxurl, data);
 	});
+
+	$(document).on('click', '.delete-google-recaptcha-keys-button', function (event) {
+		if (confirm('Would you like to permanently delete the site key and the secret key?')) {
```
andernath, line 20: Please use a textkey which can be translated

js/onoffice-handle-visibility-google-recaptcha-keys.js
```diff
@@ -0,0 +1,22 @@
+jQuery(document).ready(function ($) {
+	function showOrHiddenGoogleRecaptchaKey(inputSelector, toggleSelector) {
```
andernath, line 2: showOrHideGoogleRecaptchaKey

plugin/Renderer/InputFieldDeleteRecaptchaButtonRenderer.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
andernath, line 1: I'm not sure if this renderer is necessary.
What do you think to put this button inside of InputFieldPassworkRenderer and maybe rename it? Or do you think this is bad practice and every single element has to be a seperated class?

css/admin.css
```diff
@@ -460,8 +460,48 @@ body #post-head-content p.wp-clearfix:nth-child(1) {
 		font-size: 14px;
 		white-space: nowrap;
 	}
+
+	.oo-google-recaptcha-key {
```
andernath, line 464: media stylings should be a the end of css file to overwrite the styling on different devices. Please move the whole media block to the end and I'm sure you can remove the !important tag

andernath (commented):
@dai-eastgate I added some comments, please take a look :)

dai-eastgate:
@andernath  I will check it

dai-eastgate:
@andernath I have updated the code, can you review it again?

andernath (dismissed):
Thanks @dai-eastgate for refactoring! :)
I just got rid of the last !important tags in CSS, with a more specefic selector

fredericalpers:
On first load the css seems not to be loaded correctly. Only if I open the console and reload it is fixed. Please check and fix this behaviour.

incorrect-styling

dai-eastgate:
> On first load the css seems not to be loaded correctly. Only if I open the console and reload it is fixed. Please check and fix this behaviour.

@fredericalpers This bug occurs because the browser cache has not been cleared, so it is saving the CSS in the old branch. When you checkout to a new branch, make sure you clear your browser cache

PR #686, opened 2023-11-24
42590 Fatal Error caused by the plugin when creating menu items with "onOffice" via ACF

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/684
changed log:
Add condition check doExtraEnqueues with "onOffice" via ACF


andernath (approved):
I will test this

PR #694, opened 2023-12-06
42677 Price is emptied when property search is executed

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/679
changed log:
Change format number


plugin/Controller/InputVariableReaderFormatter.php
```diff
@@ -52,7 +52,7 @@ public function formatValue($value, string $type)
 	 */
 	public function formatFloatValue(float $value): string
 	{
-		return number_format_i18n($value, 2);
+		return number_format($value, 2, '.', '');
```
andernath, line 55: That causes problems with the format that I use to enter prices.

example in the screenshots: 

- I add one hundred (100) to search price field. 
- After searching: field formatted one hundred to 100,00 automatically. (usual format for germany)
- If I search again the price increase to ten thousand 10000,00

andernath: 2024-04-03 09_12_05-WP-Plugin Test – Testinstanz für das W3PM

andernath: 2024-04-03 09_12_29-WP-Plugin Test – Testinstanz für das W3PM

andernath (commented):
@yeneastgate please take a look to the review

plugin/Controller/InputVariableReaderFormatter.php
yeneastgate: @andernath I will check it

yeneastgate: @andernath Currently, this bug occurs for input[type=number] with fields of type float being displayed in frontend. The input display value for input[type="number"] in the frontend is being formatted by parseFloat() method and the value of input[type="number"] will vary depending on the language (Selected in WordPress Backend Site Language) 
image
**For example:**
- On the first search
-> I add one hundred (100) to the search price field. When going through the parseFloat() method , will return value = 100
-> Then when passing through the formatFloatValue() method, will return value = 100.00. Look at the following code: https://github.com/onOffice-Web-Org/oo-wp-plugin/blob/25f5a3fee9115f1fb0e05f00762c64b86b6c41ed/plugin/Controller/InputVariableReaderFormatter.php#L55
-> The value displayed in the input search price field in frontend = 100.00

- If I search again the price
->The current display value above input = 100.00, but after going through the parseFloat() method it is replaced "."  to  "". So will return value = 10000 https://github.com/onOffice-Web-Org/oo-wp-plugin/blob/25f5a3fee9115f1fb0e05f00762c64b86b6c41ed/plugin/Controller/InputVariableReaderParser.php#L111
->Then when going through formatFloatValue() method -> will return value = 10000.00

=> **This is the cause of the bug**

I have fixed it and here is video demo, please review and test this branch again. Thanks!

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/034d3cd1-44a1-4c93-b70d-734ce1bf9061

yeneastgate:
**The solution I gave above can solve this error. But I have a better solution**
As explained above, in the second search
> * If I search again the price
>   ->The current display value above input = 100.00, but after going through the parseFloat() method it is replaced "."  to  "". So will return value = 10000 https://github.com/onOffice-Web-Org/oo-wp-plugin/blob/25f5a3fee9115f1fb0e05f00762c64b86b6c41ed/plugin/Controller/InputVariableReaderParser.php#L111
>   
>   ->Then when going through formatFloatValue() method -> will return value = 10000.00

The current front-end display value above input = 100.00, but after going through the parseFloat() method has the following 
 above string processing logic
This logic is the root cause of errors when:
- The selected language in Site Language is German (or languages with $wp_locale->number_format['thousands_sep'] = '.')
- The value passed into the parseFloat() method is '100.00' after going through the above code, it will be replaced with '10000'. But the value passed to the parseFloat() method is always an int or a float containing a "." in the decimal part, so the case where the input value contains a "," will not happen => the above code is not necessary so this code can be removed

**Can you tell me what the above code does?** 
If not, I suggest removing this code, this will solve the root of the above error

andernath (approved):
Thanks @yeneastgate yeah you are right, it was the parser :)

PR #712, opened 2024-01-03
43299 settings pages property list loading time

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/700
changed log:
Change "chosen" script to "select2" script

### DIFF
--- a/css/admin.css
+++ b/css/admin.css
@@ -483,6 +483,28 @@ body #post-body-content p.wp-clearfix:nth-child(1){
 	width: 50%;
 }
 
+.oo-poststuff .custom-input-field .select2 {
+	min-width: 50%;
+}
+
+.oo-poststuff .custom-input-field .select2 .select2-selection {
+	border: 1px solid #8c8f94;
+	min-height: 30px;
+}
+
+.oo-poststuff .custom-input-field .select2 .select2-selection .select2-search--inline {
+	margin-bottom: 0;
+}
+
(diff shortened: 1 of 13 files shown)


fredericalpers:
@yeneastgate after reviewing this I came to the conclusion that we have to make a few more adjustments.

- [ ] The "x" for deselecting the items ideally should be an SVG-file e.g. [SVG Icon Close](https://css.gg/close).
- [ ] The "x" icon should be added as a <button> and must have a larger clickable area.
- [ ] Please try to avoid using floats. In this example "float: left;". :)

yeneastgate:
@fredericalpers I used SVG Icon close to replace the "x" icon and this is video evidence
Please check it and let me know your opinion

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/f278b460-5f5a-40ea-8917-e087460d3f17


> * Please try to avoid using floats. In this example "float: left;". :)

I know that we should limit the use of CSS floats as much as possible, but in this PR, I don't see myself using them. Could you point out where I went wrong?

yeneastgate:
@fredericalpers I used SVG Icon close to replace the "x" icon and this is video evidence
Please check it and let me know your opinion

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/f278b460-5f5a-40ea-8917-e087460d3f17

PR #730, opened 2024-01-22
43300 Show number of published properties

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/699
changed log:
Add button displays the number of published properties


fredericalpers:
decided against the feature because the poorer performance does not reflect the benefit.

PR #738, opened 2024-01-29
43303 Improve pagination

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/695
changed log:
Improved pagination for multiple pages


plugin/RequestVariablesSanitizer.php
```diff
@@ -42,6 +42,20 @@ public function getFilteredGet(string $name, int $filter = FILTER_DEFAULT, $opti
...
+	public function getParameterPage()
+	{
+		$pagedData = [];
+		foreach ($_GET as $key => $value) {
+			if (strpos($key, 'page_of_id_') === 0) {
+				$pagedData[] = $key;
+			}
+		}
+		return $pagedData;
+	}
```
andernath, line 58: Is this function good placed as a public function in RequestVariablesSanitizer? I think we don't need getParameterPage() for other function.
What do you think?
Should we place this code in SearchParametersModelBuilderEstate->addParametersForPagination ?

andernath:
@yeneastgate @dai-eastgate 
I finished review, please take a look :)

yeneastgate:
> @yeneastgate @dai-eastgate I finished review, please take a look :)

@andernath I will check it

plugin/RequestVariablesSanitizer.php
yeneastgate: @andernath You are correct, currently getParameterPage() is not needed for other functions. 

I placed this code in SearchParametersModelBuilderEstate->addParametersForPagination. When placing the code at this will make the code more cohesive and easier to maintain.

andernath (approved):

PR #740, opened 2024-02-01
44698 unnecessary redirects for multilingual pages with user-friendly urls

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/731
Update user-friendly URLs for multilingual pages

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -202,9 +202,13 @@ function customFieldCallback( $pDI, $format, $limitEllipsis, $meta_key ) {
 
 
 add_filter('wpml_ls_language_url', function($url) use ($pDI){
+	/** @var EstateIdRequestGuard $pEstateIdGuard */
+	$pEstateIdGuard = $pDI->get(EstateIdRequestGuard::class);
+	$pEstateDetailUrl = $pDI->get(EstateDetailUrl::class);
 	$pWPQueryWrapper = $pDI->get(WPQueryWrapper::class);
 	$estateId = (int) $pWPQueryWrapper->getWPQuery()->get('estate_id', 0);
-	return $pDI->get(EstateDetailUrl::class)->createEstateDetailLink($url, $estateId);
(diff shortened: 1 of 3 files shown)


fredericalpers:
@yeneastgate The initial issue has been resolved. However, a new issue has arisen.

There is a new redirect occurring. It appears that the URL is first being directed to the language directory with the German URL from https://w3pm.onofficeweb.com/detailseite/45-charmantes-landhaus-in-suedfrankreich/ and then redirected to the URL with the permalink in the desired language: https://w3pm.onofficeweb.com/en/detail-page/45-charming-country-house-in-the/. See attached screenshot.

Please investigate this behaviour.

redirect1

dai-eastgate:
@fredericalpers We have investigated, the cause of this 301 redirect to error is that the "objekttitel" data is obtained from API read estates call according to the current language selected on the detail page, so for switch languages ​​in the WPML, "objekttitel" will not be data according to the chosen language, it will be data of the current language on the detail page. Look at the illustration below:

image

To fix this error, the solution is to use the API to retrieve "objekttitel" according to the activated language by WPML. For each type of language option, there will be a corresponding "objekttitel"
=> I need 3 days to implement this issue. Thanks!

fredericalpers:
@dai-eastgate please go ahead and implement the recommended solution :)

dai-eastgate:
@fredericalpers Currently, I have implemented a trial version to resolve the status 301 error. 
With the current solution, it is necessary to call APIs to retrieve the object title according to the languages ​​active in the WPML Plugin.
=> Please check the solution, test it again, and let me know your opinion

Below is the test video and performance compared to master:

https://github.com/user-attachments/assets/f69981c0-d60d-4b2f-84fe-40eb9d9666f1

image

andernath (approved):

PR #744, opened 2024-02-06
44701 caching duration setting

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/720
Changed log:
Add cache duration settings feature

Issue #719, opened 2024-01-10
Hidden fields for forms


fredericalpers:
# Initiative
Improve forms of the onOffice for WP-Websites Plugin for a better user experience.

### Epic
Implement hidden fields for forms.

### User story
As a real estate agent utilizing the onOffice for WP-Websites Plugin, I encounter scenarios where certain form fields need to be hidden from customers in the frontend. This is crucial for initiating internal processes upon receiving form submissions. Presently, the plugin lacks the functionality to designate form fields as hidden. Therefore, I propose the implementation of a feature that allows me to easily mark designated form fields as hidden using a checkbox. These hidden fields should remain invisible to customers in the frontend while being transmitted both in the email notification and the attached XML file upon form submission.

The ability to set form fields as hidden is essential for streamlining internal processes triggered by form submissions. This feature is especially valuable for real estate agents who need to capture specific data for internal workflows while maintaining a seamless and uncluttered user experience on the frontend.

- [x] Implement a checkbox option in the forms to allow users to designate specific form fields as hidden. Note: GDPR status and TOS should never be marked as hidden for legal reasons.
- [x] Ensure that, upon form submission, hidden fields are included in the email sent to the designated recipient.
- [x] While the fields values that are marked with the hidden attribute  are not displayed to the user in the page's content, it is visible and can be edited using any browser's developer tools or "View Source" functionality. Therefore we also need to add the "disabled" attribute to all fields that are marked as hidden.

By successfully implementing this hidden form field functionality, the onOffice for WP-Websites Plugin can significantly enhance the flexibility and utility of forms for real estate agents.

### Further links
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/hidden
- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden
- https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

**Hidden field disabled in Backend:**
![hiddenfield1-backend](https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/88546396/6ba6400c-5828-44ef-b1b5-59072000b4de)

**Hidden field disabled in backend - Frontend display:**
![hiddenfield1-frontend](https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/88546396/35de1a60-1829-4c3c-ae81-57f264811566)

**Hidden field enabled in Backend:**
[...]

PR #745, opened 2024-02-07
44694 hidden fields for forms

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/719
changed log:
Add hidden fields feature for forms


yeneastgate:
@fredericalpers Let me test it again before you change it to in-review

fredericalpers:
Hi @yeneastgate ,

During testing, we encountered an issue where an error occurs every time a field in the form is set to "hidden" and then the fields are saved. The following error message appears: "_An error occurred while saving the form. Please check if the form name is unique._". As a result, all the fields that have been added so far are deleted, which causes the form to be empty on the frontend.

Could you please investigate and resolve this issue?

Thank you!

fehler1

yeneastgate:
@fredericalpers I will check it

yeneastgate:
@fredericalpers I apologize. 
During merging with the master branch, I missed the dbDelta function. This was the cause of the issue.

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/ff2938f1-ee47-47f4-a1bd-acfd2b63c4f5

I have now converted and fixed it. This is the video evidence. 

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/bc09cd46-a492-44d3-8e4a-2d5ace02fbc1

Could you please check and review it again?

fredericalpers:
Hi @yeneastgate ,

Thank you for addressing the previous issue. However, during further testing, we noticed a couple of additional problems:

- Hidden fields can still be marked as "Required." This should not be the case to prevent complications with empty fields in the frontend.
- Fields marked as hidden are not being declared as disabled in the frontend, although this was part of the requirements in [Issue 719 - Hidden fields for forms](https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/719).

Could you please look into these issues and make the necessary adjustments?

Thank you for your continued effort!

yeneastgate:
> * Fields marked as hidden are not being declared as disabled in the frontend, although this was part of the requirements in [Issue 719 - Hidden fields for forms](https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/719).

@fredericalpers  Do you mean that the fields selected as "hidden" will not be set as disabled like currently but they will be completely removed from the frontend, right?
image

fredericalpers:
@yeneastgate Fields marked as hidden are currently set to display: none, but they should also be disabled.

The labels of the hidden fields are still visible in the frontend. See screenshot below.

label-still-visible

yeneastgate:
@fredericalpers According to your comment, I understand that for the fields marked as "hidden", you want to remove the current "display: none" style , replace it with "disable" and show the field name as shown in the image below, is that right?
image

fredericalpers:
> @fredericalpers According to your comment, I understand that for the fields marked as "hidden", you want to remove the current "display: none" style , replace it with "disable" and show the field name as shown in the image below, is that right? image

@yeneastgate  The fields should be set to display: none; and disabeld. On top of that the labels should not be visible.

yeneastgate:
> * Hidden fields can still be marked as "Required." This should not be the case to prevent complications with empty fields in the frontend.

I updated fields marked as "hidden" cannot be marked as "required"

> * Fields marked as hidden are not being declared as disabled in the frontend, although this was part of the requirements in [Issue 719 - Hidden fields for forms](https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/719).

This feature only works for new users. For old users, please update the templates.dist folder to use this feature. You can select a template in folder templates.dist and retest because this feature is related to backward compatibility, as mentioned in the issue  https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/719#issuecomment-1929160976

This is video evidence. You can check and test it again. Thank you!

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/104892108/5cd0543f-a095-4bd4-b988-3b3f7066e333

fredericalpers:
@yeneastgate thank you, I will review this as soon as possible :)

templates.dist/form/applicantform.php
```diff
@@ -45,16 +45,20 @@
 foreach ( $pForm->getInputFields() as $input => $table ) {
```
andernath, line 45:
```suggestion
foreach ( $pForm->getInputFields() as $input => $table ) {
    if($pForm->isHiddenField($input)) {
        $hiddenValues []= renderFormField($input, $pForm);
        continue;
    }
```

andernath: To do not add ```<br>``` I would use an hidden value array

In line 82 and 88 I would add this array
```<?php echo implode( $hiddenValues); ?>```

File: templates.dist/form/ownerform.php [hunk 2]
```diff
@@ -45,22 +45,29 @@
 	foreach ( $pForm->getInputFields() as $input => $table ) {
 		$isRequired = $pForm->isRequiredField($input);
```
andernath, line 46: Please see suggestion in applicantform to remove <br> on hidden fields

templates.dist/form/ownerleadgeneratorform.php
```diff
@@ -52,7 +52,9 @@
```
andernath, line 52: Please see suggestion in applicantform to remove <br> on hidden fields

andernath (changes requested):
@yeneastgate please take a look :)

yeneastgate:
@andernath I will check it asap

yeneastgate:
@andernath 

> Is this JS really needed?
> 
> What kind of nextSibling do you remove here? Maybe you can add a check in form template, to decide to render this sibling or not.
> 
> This seems possible to me with nextBr. See my suggestions in defaultform.php and applicantform and tell me if you agree? :)

If we use the solution you suggested, we can remove this js file. Thank you!

> I would prefer naming: $line = $label and echo without `<br>` if field is hidden. See suggestion

> I would prefer naming: $line = $label
> and echo without 'br' if field is hidden. See suggestion

I have updated it

> This can never be hidden or??

You are right, I already have the check condition in admin.js so there is no need to add code in here

> To do not add `<br>` I would use an hidden value array
> 
> In line 82 and 88 I would add this array `<?php echo implode( $hiddenValues); ?>`

Placing this code on lines 82 and 88 will cause duplicate code. See the image below
image

I think we should put them after the "Your search criteria" rendering section.

andernath:
> Placing this code on lines 82 and 88 will cause duplicate code. See the image below
@yeneastgate  Ops 😁 yeah your right,.. Below search criteria is a good place 👍

andernath:
@yeneastgate 
For you information: I merged https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/744 with database version changes into master.
You have to merge this. :)

Please tell us if this pull request is ready to review again.

By the way: I will be on vacation for the next 2 weeks. 😎

yeneastgate:
> @yeneastgate For you information: I merged #744 with database version changes into master. You have to merge this. :)
> 
> Please tell us if this pull request is ready to review again.
> 
> By the way: I will be on vacation for the next 2 weeks. 😎

@andernath I will check and test the branch and let you know as soon as possible. Have a great vacation! 😊

css/onoffice-forms.css
```diff
@@ -9,4 +9,8 @@ span.onoffice-pleasefill {
 	opacity: 0.5;
 	cursor: not-allowed;
 	pointer-events: none;
+}
+
+.oo-form .hidden-field {
+	display: none;
```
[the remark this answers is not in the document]
yeneastgate: @andernath  We should keep the parent class "oo-form" because some themes will have their own CSS styles to customize input fields. If you don't use the parent class, the priority of the above style snippet will be lower, and then the attribute "display:none" will not be executed.

templates.dist/fields.php
```diff
@@ -122,6 +122,7 @@ function renderFormField(string $fieldName, onOffice\WPlugin\Form $pForm, bool $
 		$selectedValue = $pForm->getFieldValue($fieldName, true);
 		$isRangeValue = $pForm->isSearchcriteriaField($fieldName) && $searchCriteriaRange;
 		$fieldLabel = $pForm->getFieldLabel($fieldName, true);
+		$isHiddenField = $pForm->isHiddenField($fieldName);
```
pospisk, line 125: Please define the hidden attribute here 
using:
```
$isHiddenField = $pForm->isHiddenField($fieldName);
$hiddenAttribute = $isHiddenField ? 'class="oo-hidden-field" readonly ' : '';
```

instead of:
```
if ($isHiddenField === true) {
	$inputType .= 'class="oo-hidden-field" disabled ';
}
```

templates.dist/fields.php
```diff
@@ -172,14 +175,16 @@ function renderFormField(string $fieldName, onOffice\WPlugin\Form $pForm, bool $
...
+			$output .= $isHiddenField ? 'class="oo-hidden-field" disabled ' : 'class="custom-multiple-select form-control"' . '>';
 			$output .= $htmlOptions;
 			$output .= '</select>';
 		} else {
 			$inputType = 'type="text" ';
 			$value = 'value="' . esc_attr($pForm->getFieldValue($fieldName, true)) . '"';
-
-			if ($typeCurrentInput == onOffice\WPlugin\Types\FieldTypes::FIELD_TYPE_BOOLEAN) {
+			if ($isHiddenField === true) {
+				$inputType .= 'class="oo-hidden-field" disabled ';
```
pospisk, line 186: Please remove:
```
if ($isHiddenField === true) {
	$inputType .= 'class="oo-hidden-field" disabled ';
} else
```

to restore the original if statement:
```
if ($typeCurrentInput == onOffice\WPlugin\Types\FieldTypes::FIELD_TYPE_BOOLEAN) {
```

afterwards to make it work please insert the following `$inputType .= $hiddenAttribute;` around line 210 

```
} elseif (
	$typeCurrentInput === FieldTypes::FIELD_TYPE_DATETIME
) {
	$inputType = 'type="datetime-local" step="1" ';
}

// here

if (
	$isRangeValue && $pForm->inRangeSearchcriteriaInfos($fieldName) &&
	count($pForm->getSearchcriteriaRangeInfosForField($fieldName)) > 0
) {
```

Bug:
Currently when a checkbox is set to hidden the input type="" is stuck on "text" when it should be "checkbox". This causes an incorrect value in the field and they aren't submitted in the form. 

hidden

templates.dist/fields.php
```diff
@@ -214,7 +219,8 @@ function renderFormField(string $fieldName, onOffice\WPlugin\Form $pForm, bool $
 						. $value . ' placeholder="' . esc_attr($rangeDescription) . '">';
 				}
 			} elseif ($typeCurrentInput === FieldTypes::FIELD_TYPE_DATATYPE_TINYINT) {
-				$output = '<fieldset>
+				$additionHidden = $isHiddenField === true ? 'class="oo-hidden-field" disabled ' : '';
+				$output = '<fieldset '.$additionHidden.'>
```
pospisk, line 223: Please remove `$additionHidden`, you can use `$hiddenAttribute` instead which is defined at line 126.

templates.dist/form/defaultform.php
```diff
@@ -65,16 +65,22 @@
...
-			echo $pForm->getFieldLabel( 'message' );
-			echo $additionMessage . ':<br>';
-			echo '<textarea name="message">' . $pForm->getFieldValue( 'message' ) . '</textarea><br>';
+			$isHiddenField = $pForm->isHiddenField('message');
+			if (!$isHiddenField) {
+				echo $pForm->getFieldLabel( 'message' );
+				echo $additionMessage . ':<br>';
+				echo '<textarea name="message">' . $pForm->getFieldValue('message') . '</textarea><br>';
+			} else {
+				echo '<textarea name="message" class="oo-hidden-field" disabled">' . $pForm->getFieldValue('message') . '</textarea>';
```
pospisk, line 74: There is a stray `"` double-quotation after `disabled` and please use `readonly` instead of `disabled`.

templates.dist/form/ownerform.php
```diff
@@ -58,9 +63,14 @@
...
-	
-			$line = $pForm->getFieldLabel( 'message' ).$additionMessage.':<br>';
-			$line .= '<textarea name="message">'.$pForm->getFieldValue('message').'</textarea><br>';
+			$isHiddenField = $pForm->isHiddenField('message');
+			if (!$isHiddenField) {
+				$line = $pForm->getFieldLabel( 'message' );
+				$line .= $additionMessage . ':<br>';
+				$line .= '<textarea name="message">' . $pForm->getFieldValue('message') . '</textarea><br>';
+			} else {
+				$line = '<textarea name="message" class="oo-hidden-field" disabled">'.$pForm->getFieldValue('message').'</textarea><br>';
```
pospisk, line 72: There is a stray `"` double-quotation after `disabled` and please use `readonly` instead of `disabled`.

pospisk (changes requested):
@yeneastgate 
Please take a look at the requested changes.

yeneastgate:
@pospisk I have refactored the code, and after checking, I found that in some cases, such as select elements, the readonly attribute cannot be applied. Therefore, I have added the disabled attribute for those instances. You can see this in the video.

https://github.com/user-attachments/assets/26826467-b67a-4128-acd8-6f314a46e17b


Please review the changes and let me know your opinion.

plugin/FormPost.php
```diff
@@ -227,6 +231,42 @@ private function buildFormData(DataFormConfiguration $pFormConfig, $formNo): For
...
+		foreach (array_chunk($fields, 100) as $fields) {
+			$pDefaultFields = $pDefaultValueRead->getConvertedMultiFields($formId, $fields);
+			if (count($pDefaultFields)) $values = array_merge($values, $pDefaultFields);
+		}
+
+		array_walk($values, function($value, $key) {
+			if ($value === "0" || !empty($value)) {
+				$_POST[$key] = $value;
+			}
+		});
```
pospisk, line 268: form

**Issues:** 
- If a field has a default value set, we can remove it from the HTML but it will still be submitted.
- Changing the value in HTML can’t overwrite the default value set.
- Hidden checkboxes with a default value "y" submit "1" which is interpreted as "n".

**Fix:**
_plugin/FormPost.php_

```
private function readDefaultValueForHiddenFieldsByFormId(int $formId, array $hiddenFields)
...
array_walk($values, function($value, $key) {
	if ($value === "0" || !empty($value) && !isset($_POST[$key])) {
		$_POST[$key] = $value;
	}
});
```
We need `!isset($_POST[$key])` to make sure we can overwrite the default value set in the plugin

pospisk (changes requested):
The fix within FormPost.php requires changes in the fields.php and not all parts of the file are visible within the review. (see next message)

yeneastgate:
@pospisk I will check it.

yeneastgate:
@pospisk Based on the issues and solutions you provided, it seems that we can still edit the HTML by using the browser's developer tools or "View Source". However, according to the initial requirements, you want us to disable this behavior. https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/719#issue-2074416434
> While the fields values that are marked with the hidden attribute are not displayed to the user in the page's content, it is visible and can be edited using any browser's developer tools or "View Source" functionality. Therefore we also need to add the "disabled" attribute to all fields that are marked as hidden.

Please confirm the following:
- Editing HTML using the browser's developer tools or "View Source".
- If a field is marked as hidden, its type should be changed to hidden.

yeneastgate:
@pospisk Based on the issues and solutions you provided, it seems that we can still edit the HTML by using the browser's developer tools or "View Source". However, according to the initial requirements, you want us to disable this behavior. https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/719#issue-2074416434
> While the fields values that are marked with the hidden attribute are not displayed to the user in the page's content, it is visible and can be edited using any browser's developer tools or "View Source" functionality. Therefore we also need to add the "disabled" attribute to all fields that are marked as hidden.

Please confirm the following:
- Can we edit the HTML using the browser's developer tools or "View Source" and have the modified values submitted instead of the default values?
- If a field is marked as hidden, do you want all types to be changed to hidden?

fredericalpers:
> @pospisk Based on the issues and solutions you provided, it seems that we can still edit the HTML by using the browser's developer tools or "View Source". However, according to the initial requirements, you want us to disable this behavior. [#719 (comment)](https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/719#issue-2074416434)
> 
> > While the fields values that are marked with the hidden attribute are not displayed to the user in the page's content, it is visible and can be edited using any browser's developer tools or "View Source" functionality. Therefore we also need to add the "disabled" attribute to all fields that are marked as hidden.
> 
> Please confirm the following:
> 
> * Can we edit the HTML using the browser's developer tools or "View Source" and have the modified values submitted instead of the default values?
> * If a field is marked as hidden, do you want all types to be changed to hidden?

Please go ahead and implement @pospisk suggested solution. 

> * Can we edit the HTML using the browser's developer tools or "View Source" and have the modified values submitted instead of the default values?
> * If a field is marked as hidden, do you want all types to be changed to hidden?

I can confirm both :)

yeneastgate:
@fredericalpers  @pospisk I have updated the code and here is evidence https://files.fm/f/sdeeqnrw9h
Please review and recheck it. Thanks!

pospisk (approved):
Excellent, thank you very much.

PR #762, opened 2024-02-28
Fix slick style order in release v4.18

andernath:
#4127913


ogzA:
I've tested the changes and can confirm the issue is resolved. Everything looks solid, great work!

Issue #235, opened 2022-05-04
P#71173 Make CSS local to templates


jmaas-onoffice:
# Current state
Changes to the CSS files (especially [`onoffice_defaultview.css`](https://github.com/onOfficeGmbH/oo-wp-plugin/blob/5109fa67edaa2256f2ed5357d503ba3009ea32b5/css/onoffice_defaultview.css)) can break personalized templates. Since `onoffice_defaultview.css` also gets loaded for personalized templates, users may rely on it. If we change it, we can only test it with the default templates, so it could break users' templates.

This means we do not want to touch those CSS files and cannot make improvements to the included templates.

# Desired state
1. The CSS should be coupled to the templates. We can achieve this by loading a CSS file from the templates folder. That path should be `templates/onoffice-style.css`. (This allows users to copy the CSS alongside the templates. We can then change the included CSS, because it will not affect the copied CSS file.)

    The possible paths:
    - Included: `wp-content/plugins/onoffice-for-wp-websites/templates.dist/onoffice-style.css`
    - Plugin folder: `wp-content/plugins/onoffice-personalized/templates/onoffice-style.css`
    - Theme folder: `wp-content/themes/<theme>/onoffice-theme/templates/onoffice-style.css`

2. The change should be backwards compatible, so that users can keep their existing templates without changes.

    We use the existing template folder hierarchy (theme, then plugin, then included). Look in the highest existing template folder. If there is a `onoffice-style.css`, use it. Otherwise use the old `onoffice_defaultview.css`.
  
    Cases:
    - If there is a theme folder:
      - If it contains a `onoffice-style.css`, use that.
      - If it does not contain a `onoffice-style.css`, use the old `onoffice_defaultview.css`.
    - Else, if there is a plugin folder:
      - If it contains a `onoffice-style.css`, use that.
      - If it does not contain a `onoffice-style.css`, use the old `onoffice_defaultview.css`.
    - Else (there are only the included templates):
      - Use the included `wp-content/plugins/onoffice-for-wp-websites/templates.dist/onoffice-style.css`.

LongTrong-exe:
@jmaas-onoffice Your examples is so confusing :confused::cry:

PR #765, opened 2024-03-07
45503 implement AreaButler in detail page template

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/748
changed log:
Add AreaButler on detail page template


css/onoffice-default.css
```diff
@@ -71,4 +71,9 @@ form[data-applicant-form-id] #spinner {
 
 .message{
     display: none!important;
+}
+
+.oo-area-butler {
+    position: relative;
+    height: 500px;
 }
```
andernath, line 79: I think this fits more in onoffice-defaultview.css
Whats your opinion?

andernath (commented):
@dai-eastgate Please take a look at my suggestion

dai-eastgate:
@andernath I will check it

dai-eastgate:
@andernath 
- I found that the issue https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/235 mentioned not adding CSS to the onoffice_defaultview.css file because it can lead to the breaking of personalized templates.
=> So please check the information before you want to move the CSS to the onoffice_defaultview.css file. If you still want to update the CSS in this file, I will fix it.
- Please explain the disadvantages of placing CSS in onoffice-default.css

andernath:
> * I found that the issue [P#71173 Make CSS local to templates #235](https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/235) mentioned not adding CSS to the onoffice_defaultview.css file because it can lead to the breaking of personalized templates.
>   => So please check the information before you want to move the CSS to the onoffice_defaultview.css file. If you still want to update the CSS in this file, I will fix it.

Oh Thanks for investigating! I see we should not use onoffice_defaultview.css neither onoffice-default.css
Instead of template.dist/onoffice-style.css seems to be the right one. Between /* details */ and /* responsive */ codeblock.

> * Please explain the disadvantages of placing CSS in onoffice-default.css

I got suspicious because of the file naming and classes inside. But anyway, thanks to your hint, we've found out where we need to insert styling changes: 
template.dist/onoffice-style.css


Don't worry. **I will add this change.** So Frederic can start testing. :)

andernath (approved):

dai-eastgate:
@andernath For this new AreaButler feature, placing the CSS in template.dist/onoffice-style.css makes sense, but for some old features, be careful placing the CSS in template.dist/onoffice-style.css because this may affect backward compatibility

fredericalpers:
@dai-eastgate After reviewing there are a few changes that have to be made.

- [ ] Add a H2-Headline with the text `AreaButler`
- [ ] Add styling as for the other sections `padding-bottom: 15px; padding-top: 15px; border-bottom: 2px solid #efefef;`

dai-eastgate:
> * [x]  Add a H2-Headline with the text `AreaButler`
> * [x]  Add styling as for the other sections `padding-bottom: 15px; padding-top: 15px; border-bottom: 2px solid #efefef;`

@fredericalpers I have updated CSS. This is video demo:

https://github.com/onOffice-Web-Org/oo-wp-plugin/assets/106214469/8f8fec0e-1245-403c-9e6f-c3051b40cbe5

Note: Currently I cannot access https://areabutler.dev.areabutler.de/ on the onOffice Enterprise Enterprise so the map for AreaButler is not displayed, instead it is showing as "loading"
Please check and let me know your opinion. Thanks!

andernath (approved):

PR #783, opened 2024-03-22
46121 change incorrect label text for address list

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/777
changed log:
Change incorrect text label for address list


miauzjpg (approved):
good job! 👍

PR #788, opened 2024-03-28
46248 resolve fatal error caused by price on request with similar estate

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/786
changed log:
Fix bug price on request for similar estates


miauzjpg (approved):
good job 🎉

PR #796, opened 2024-04-10
Add classes for custom styling

andernath:
#4052783


andernath:
Default page shows a helper if no estate can be displayed, because of wrong url. This helber is now able to be styled with customer theme. Example: image

fredericalpers (approved):

PR #800, opened 2024-04-22
47254 Files are not loaded on property detail page

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/799
changed log:
Load script for unit


miauzjpg (approved):
good job 👍

PR #823, opened 2024-05-29
47945 Improvement of the address list functionality - Add property list selection

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/808
changed log:
Add property list selection config in address detail


andernath:
Merged to master with https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/889

PR #821, opened 2024-05-24
47946 Optional writing of activities in the agents log with plugin forms & fields.

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/809
changed log:
Add activity option & create agents log when create forms

PR #826, opened 2024-05-31
48659 Optional writing of tasks in onOffice enterprise with plugin forms & fields.

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/820
changed log:
Add task option & create task when create forms


fredericalpers:
@yeneastgate Thank you for the work so far! The task creation functionality from the forms is working well. However, there are a few outstanding issues that need to be addressed:

**Missing Task Type "Callback":**
The task type "Callback" is not yet available. This needs to be added to ensure that users can select this type when creating tasks. Please implement this as part of the task creation process.

**Field Enable/Disable Logic:**
Currently, the fields related to task creation should be disabled when the checkbox for creating tasks is not selected. However, it appears that this logic isn't functioning as expected. When the checkbox is unchecked, all other fields should be disabled to prevent any accidental input. Please update the UI to reflect this behavior.

**Linking Tasks to Incoming Emails:**
To provide better context for tasks, they should be linked to the incoming emails from which they originate. This can be achieved by using the mailMessageId parameter, which allows the task to be associated with the specific email. Refer to the [onOffice API documentation for more details](https://apidoc.onoffice.de/actions/datensatz-anlegen/tasks/) on how to implement this. Ensuring that tasks are properly linked to emails will improve traceability and context.

Once these issues are addressed, the feature will be much closer to completion. Please make the necessary adjustments and let me know if you have any questions. Thanks again for your efforts

yeneastgate:
@fredericalpers I will check it.

fredericalpers:
@yeneastgate please ignore the connection to the email.

For now only check the field enable/disable logic which also needs implementation in #821 

Thank you :)

fredericalpers:
Vorgezogen auf 5.2

yeneastgate:
> **Missing Task Type "Callback":**
> The task type "Callback" is not yet available. This needs to be added to ensure that users can select this type when creating tasks. Please implement this as part of the task creation process.

@fredericalpers I found that the "callback" API with the key "Rückruf" currently does not return a label. Can you check with the IT department? Otherwise, we will hard code it, but it will not be translatable.

image

fredericalpers:
> > **Missing Task Type "Callback":**
> > The task type "Callback" is not yet available. This needs to be added to ensure that users can select this type when creating tasks. Please implement this as part of the task creation process.
> 
> @fredericalpers I found that the "callback" API with the key "Rückruf" currently does not return a label. Can you check with the IT department? Otherwise, we will hard code it, but it will not be translatable.


@yeneastgate please hard code it for now, using a onOffice Textkey, then we should be able to translate it in WordPress. Thank you. :)

yeneastgate:
@fredericalpers I got it.

yeneastgate:
> **Missing Task Type "Callback":**
> The task type "Callback" is not yet available. This needs to be added to ensure that users can select this type when creating tasks. Please implement this as part of the task creation process.

@fredericalpers I fixed it. Please take a look at my video and let me know your opinions. Thanks!

https://github.com/user-attachments/assets/2b0e0b6d-69a7-41f7-b47a-257dacdc6674

fredericalpers:
@dai-eastgate It appears that there are merge conflicts in this branch that need to be resolved before we can proceed with the merge. Please resolve the conflicts in the affected files and update the branch. Thank you in advance :)

andernath (approved):

PR #846, opened 2024-06-28
50189 reflected XSS

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/844
changed log:
Update code to prevent Cross-Site Scripting (XSS)


ingoKrOffice (commented):
passt soweit

Issue #847, opened 2024-07-01
Displaying multiple property lists and OSM on one page - Fix after leaflet.js update


fredericalpers:
# Displaying multiple property lists and OSM on one page - Fix after leaflet.js update

### Introduction
As mentioned in Issue https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/703#issuecomment-1907305962 there is still an issue that needs to be addressed.

### User Story
As a user of the onOffice WordPress plugin, I want each property list and OSM map on a single page to display the correct properties and maps, so that all information is correctly and clearly presented.

Following the release of the Improve Pagination https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/695 update, an issue has arisen when multiple property lists and OSM maps are used on the same page. The current state results in an error and no OSM for the second property list. Thus, an additional fix is necessary to ensure that each property list matches the correct OSM map and displays the correct properties.

<details><summary>Error or also see browser console</summary>
<p>

Uncaught Error: Map container is already initialized.
    at e._initContainer (Map.js:1094:10)
    at e.initialize (Map.js:136:8)
    at new e (Class.js:24:20)
    at t.map (Map.js:1750:9)
    at immobilienliste-alle-immobilien/:539:21
    at immobilienliste-alle-immobilien/:557:7

</p>
</details> 

### Task / Requirements

- [x] Fix the bug mentioned in https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/703#issuecomment-1907305962
- [x] Each property list must display a map (if activated and one of the propertys address is shared)
- [x] Each  map must show the correct locations of properties that are displayed in the corresponding list.
- [x] Ensure that each OSM map displays only the locations of the properties in the corresponding list.
- [x] Regression tests to ensure that previous functionalities are not affected.

yeneastgate:
@fredericalpers  I have fixed this bug. Please check the link video evidence below:
- For single property list: https://files.fm/u/yku75ynzp7#/view/n5njun8nn7
- For Multiple property lists:  https://files.fm/u/kzv7j7x6p4#/view/kgj2d7ajt4

**Note:**
1. Only new users can use this feature. For old users, please copy the template.dist folder to use this feature.
2. I also encounter the same bug when using Google Maps
Please check and let me know your opinion. Thanks!

fredericalpers:
@yeneastgate Seems to be working for mutliple lists and OSM now, thank you :)

[...]

PR #854, opened 2024-07-03
50475 Displaying multiple property lists and OSM on one page - Fix after leaflet.js update

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/847
changed log:
Fix bug displaying multiple property lists and OSM on one page


fredericalpers:
Vorgezogen auf 5.2

fredericalpers:
I've noticed an issue when multiple property lists are added to the page.

- The styling for the height of the map in the first property list seems to be ignored. Unlike the second map, which renders with the correct dimensions, the first map does not maintain its intended height. This leads to an inconsistent user experience, with one map appearing correctly and the other not.
- Another issue that arises is that the zoom level of the first map defaults to being completely zoomed out (100%). This is problematic because the zoom level should be consistent across both maps, and ideally, the first map should be initialized with the same zoom level as the second one to ensure uniformity across the property lists.

It’s worth investigating whether there has been any change in the class or ID associated with the first map container. If a class name has been modified or is missing, it could explain why the height styling is not being applied properly. It’s also possible that the map initialization script is targeting the wrong element or applying the necessary styles only to the second map, leaving the first map in its default state.

yeneastgate:
@fredericalpers We will check and fix it

yeneastgate:
@fredericalpers Please provide a specific image, video, or link so we can check.

Upon testing, I found that all maps' default height is 300px and the zoom level between maps is consistent.

https://github.com/user-attachments/assets/f0207a24-3cc8-4337-95f4-855b43d25d2f

For instance, if the locations of the properties are far apart, the map will automatically zoom out to fully display all locations. In particular, the map will adjust the zoom level based on the distance between the two furthest points. This ensures that all marked locations are displayed within the same frame.

If we remove the fitBounds method, the map will not automatically adjust the zoom level to fit all markers within the view. This means that some markers might not be visible if they are outside the initial zoom level and center of the map.

https://github.com/user-attachments/assets/ae10ddf6-622c-4d39-9ad0-f3ed386c4705

fredericalpers:
Hi @yeneastgate ,

Unfortunately, the feature isn't working as expected. I added two property lists, both have the map enabled, and both include properties where the addresses are visible. However, the map is only displayed for the property list that uses the template from the personalized folder. In the list using the default template, the map's CSS is missing.

Additionally, when both lists use the template from the personalized folder, the map is only displayed for the first property list.

You can reproduce the issue on the following page: [https://testinstanz-fal.onofficeweb.com/immobilienliste-alle-immobilien/](https://testinstanz-fal.onofficeweb.com/immobilienliste-alle-immobilien/)

Could you please take another look?

yeneastgate:
@fredericalpers This issue requires users to update the template as I mentioned earlier in this comment https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/847#issuecomment-2205772547. Since the issue is related to the template, I'm not sure if older users can use it without updating the template.
Could you please try updating the template and check if the issue is resolved?

fredericalpers:
> @fredericalpers This issue requires users to update the template as I mentioned earlier in this comment [#847 (comment)](https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/847#issuecomment-2205772547). Since the issue is related to the template, I'm not sure if older users can use it without updating the template. Could you please try updating the template and check if the issue is resolved?

When testing I always deactive the current plugin and rename the folder, so its not in use at all. Then I install the PRs Zip. The template.dist therefore should be the new one with all relevant files and templates. Maybe theres misunderstanding. Even with two default templates this doesn't work yet.

yeneastgate:
@fredericalpers After checking, I found that your first shortcode is using template.dist and the second shortcode is using onoffice-personalized.
image

Currently, the plugin only activates one style (onoffice-style), and the priority of onoffice-personalized is higher than template.dist. Therefore, the page is using onoffice-style from onoffice-personalized.

There is no new style in the onoffice-style file yet!

image

Could you update the onoffice-style in onoffice-personalized and check again? You can watch the video for more details.
https://files.fm/f/eycchgwnn3

fredericalpers:
@yeneastgate Currently the feature with the cards for 2 lists only works if you have the new styles file in the Personalized folder.

I have selected the default templates, but it still seems to pull the style file from the personalized folder. This must not be the case, otherwise the feature will not work if I do not have a personalized folder and only use the default of the plugin.

yeneastgate:
@fredericalpers I understand what you are saying. Currently, the plugin's mechanism is that if there is an onoffice-personalized folder and it contains the onoffice-style file, it will automatically prioritize and select it. Even if you choose a template in the default plugin, it will still prioritize the onoffice-personalized folder if it exists.

For more details, you can refer to the discussion on GitHub Issue #235.

fredericalpers:
@yeneastgate thanks for explaining and clearing up the confusion! :)

andernath (approved):

PR #858, opened 2024-07-10
50561 Proper Handling of Image Type Selection in Property List Settings

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/851
changed log:
Fix bug image type selection in the property list


miauzjpg (approved):
good!

Issue #799, opened 2024-04-17
Files are not loaded on property detail page


fredericalpers:
# Files are not loaded on property detail page

### Introduction
In version 4.18 of the onOffice for WP-Websites plugin, a bug has occurred that causes certain files not to be loaded when the [oo_estate units="units" view="detail"] shortcode is used.

**Steps for replication:**

1. Install version 4.18 or 4.18.1 of the onOffice for WP-Websites plugin.
2. Save units shortcode e.g. [oo_estate units="units" view="detail"] in a page.
3. Open the detail page of a property.

**Expected behavior:**
The files like onoffice_detailview-js, slick.js and slick.ss should be loaded to support the functionality of the onOffice plugin.

**Observed behavior:**
The mentioned files are not loaded, which leads to an incorrect display of the slick slider.

### User Story
As a real estate agent using the onOffice for WP Websites plugin, I expect all relevant files to load properly when I use the units shortcode to display units as well.

This is important for my website to function correctly and for potential clients to have a smooth experience when searching for properties and units. Failure to load these files can result in an incorrect display, which in turn affects customer confidence in my website and service.

By fixing the bug, the impaired function of the onOffice for WP website should be restored. This will result in a better user experience and a professional presentation for potential customers.

### Requirements / Tasks

- [x] Analysis of the code to identify the cause of the bug.
- [x] Adjustment of the code so that the required files are loaded.
- [x] Conduct thorough tests and avoid possible regressions.

PR #861, opened 2024-07-12
50844 Fix Email sending and creating data records issue with Honeypot Function on Property Detail Pages

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/857
changed log:
Add condition load honeypot on detail page when using ACF


yeneastgate:
@fredericalpers @andernath I have checked and found that before WordPress version 5.2, the get_the_content function did not work correctly when retrieving the content of a post, which prevented shortcodes from being fetched and scripts from being loaded accordingly.
I have made some changes to use an optimal approach because this is a core attribute in WordPress.
Here is video evidence wordpress version 5.1.19:

https://github.com/user-attachments/assets/13c246b9-b09a-41cf-9c97-6be4d3990f63

fredericalpers:
@yeneastgate Our further tests have shown that:

- The honeypot field is not added to the form at all. Therefore no e-mails are received.
- The problem ONLY seems to exists if the units are to be displayed in the shortcode ([oo_estate units="Units list" view="detail"]). 
- It also does not seem to work if the shortcode with unit list is stored via the Gutenberg "Shortcode".

https://testinstanz-mr.onofficeweb.com/
Here the suggested fix also works for the unit list with the current detail page. As soon as a new detail page is created, the whole functionality with units no longer works. We suspect that the shortcode was stored somewhere outside ACFs/Gutenberg in the past, e.g. in the Classis Editor or in the database.

Please investigate further. Thank you :)

yeneastgate:
@fredericalpers You are right, this issue is related to issue #799,
=> I merged PR #799 into this PR and tested it again using the editors: Classic Editor, Advanced Custom Fields, and Gutenberg editor, and tested the unit detail functionality, and it still works fine.

yeneastgate:
@fredericalpers You are right, this issue is related to issue https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/799,
=> I merged PR https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/799 into this PR and tested it again using the editors: Classic Editor, Advanced Custom Fields, and Gutenberg editor, and tested the unit detail functionality, and it still works fine.

fredericalpers:
@yeneastgate Please double check if you merged the PR's. 

I can not find a commit for it, therefore I'm not able to deploy the build to test everything :)

yeneastgate:
> @yeneastgate Please double check if you merged the PR's.
> 
> I can not find a commit for it, therefore I'm not able to deploy the build to test everything :)

@fredericalpers I have merged it. You can deploy and test it now.

fredericalpers:
This seems to be fixing the problem :) thank you

fredericalpers:
https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/857#issuecomment-2232555992

fredericalpers:
Bereits behoben mit https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/800

PR #869, opened 2024-07-26
51413 Critical error when using empty "Radius" for Similar Estates

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/859
changed log:
Changed HTML input type from `text` to `number` for radius and record per page


fredericalpers:
@dai-eastgate Thank you for the update, it's working well now that no error is shown for an empty radius value! 🎉

I do have a small improvement suggestion regarding the radius field: it should not be possible to enter a negative value. Currently, it's still possible to input negative numbers, which probably doesn’t make sense in this context. Could you add a validation to prevent negative values?

Thanks in advance!

dai-eastgate:
@fredericalpers I fixed it. Please help us to review it. Thanks!  

https://github.com/user-attachments/assets/02d13838-b7bb-4204-9f45-e9cc850d6ab1

fredericalpers:
@dai-eastgate thank you, I will reviews this as soon as possible

miauzjpg (approved):
good job

PR #909, opened 2024-09-26
53099 Integration of "City-Select" into the Location Field of the Address List Search

yeneastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/908
changed log:

- Added a checkbox in the backend to enable display as a dropdown.
- When enabled, the location field in the address list search is presented as a dropdown menu instead of a text input.


andernath (approved):
Good work 👍 @yeneastgate 
I just refactored one little phpdoc param :)

PR #923, opened 2024-10-08
53444 Implementation of the “bildWebseite” field in adresses

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/914
changed log:
update <code>bildWebseite</code> field in address list & address detail


plugin/AddressDetail.php
```diff
@@ -56,6 +57,11 @@ public function loadSingleAddress($id)
         if($this->getDataViewAddress() instanceof DataAddressDetailView &&
             $this->getDataViewAddress()->getPictureTypes() != [] && $this->getDataViewAddress()->getPictureTypes() != null)
             array_push($fields, "imageUrl");
+
+		if($this->getDataViewAddress() instanceof DataAddressDetailView && $this->getDataViewAddress()->getPictureTypes() != null && 
+			in_array(ImageTypes::BILDWEBSEITE, $this->getDataViewAddress()->getPictureTypes()))
+			array_push($fields, ImageTypes::BILDWEBSEITE);
+
```
andernath, line 64: What about combining both If conditions?

andernath (commented):
@dai-eastgate Just one question everything else is fine :)

plugin/AddressDetail.php
dai-eastgate: @andernath You're right, we should combine them. I've made the changes. Thank you!

andernath (approved):

PR #940, opened 2024-10-23
Fix missing social meta data settings

andernath:
Different settings shared same groupslug and overwrite eachother


plugin/Gui/AdminPageApiSettings.php
```diff
@@ -617,7 +617,7 @@ private function addFormModelCache()
 	 */
 	private function addFormModelSeparatorFormatSettings(string $pageSlug)
 	{
-		$groupSlugPaging = 'onoffice-settings';
+		$groupSlugPaging = 'onoffice-thousand-separator';
```
dai-eastgate, line 620: I think we should only change the GroupSlug in addFormModelSocialMetaData. If we change it in addFormModelSeparatorFormatSettings, users who have saved settings might need to save the SeparatorFormat settings again.

plugin/Gui/AdminPageApiSettings.php
```diff
@@ -617,11 +617,10 @@ private function addFormModelCache()
 	 */
 	private function addFormModelSeparatorFormatSettings(string $pageSlug)
 	{
-		$groupSlugPaging = 'onoffice-thousand-separator';
 		$labelSeparatorFormatSettings = __('Separator Format Settings', 'onoffice-for-wp-websites');
 		$labelSeparatorCharacterFormat = __('Thousand Separator Format', 'onoffice-for-wp-websites');
 
-		$pInputModelSeparatorCharacterFormat = new InputModelOption($groupSlugPaging, 'thousand-separator',
+		$pInputModelSeparatorCharacterFormat = new InputModelOption('onoffice-settings', 'thousand-separator',
```
andernath, line 633: @dai-eastgate 👍 Good Job! :)

PR #957, opened 2024-11-07
54399 Correct handling of Umlauts and Special Characters during language switch

dai-eastgate:
related to https://github.com/onOffice-Web-Org/oo-wp-plugin/issues/953
changed log:


plugin/Controller/EstateDetailUrl.php
```diff
@@ -35,6 +35,7 @@ class EstateDetailUrl
 	 * @param string|null $title
 	 * @param string|null $oldUrl
 	 * @param bool $flag
+	 * @param string|null $switchLocale
```
andernath, line 38: Is type null needed?

andernath (commented):
@dai-eastgate 
Everything looks fine :), just one question. Could you please take a look?

plugin/Controller/EstateDetailUrl.php
dai-eastgate: @andernath It really isn't necessary. I'll fix it.

PR #965, opened 2024-11-27
54907 Empty div in the template of the detail page

dai-eastgate:


fredericalpers:
Fixed in https://github.com/onOffice-Web-Org/oo-wp-plugin/pull/1010

PR #989, opened 2025-01-17
Bug/T#4646119-special-char-search

mmilawski:
Decodes a html string for filtering. In following case there is an issue when strings contain special characters like apostrophes, which are encoded as `&#039;` and is passed along to the oO API. For example if the user is searching for `Cortina d'Ampezzo`, the string is encoded to `Cortina d&#039;Ampezzo`. The API will not find any items because the string seems to not be decoded in the enterprise API.

### DIFF
--- a/plugin/Filter/FilterBuilderInputVariables.php
+++ b/plugin/Filter/FilterBuilderInputVariables.php
@@ -102,9 +102,9 @@ class FilterBuilderInputVariables
 	private function getFieldFilter($fieldValue, string $type): array
 	{
 		$fieldFilter = [];
-        if(is_string($fieldValue)){
-            $fieldValue = html_entity_decode($fieldValue);
-        }
+		if(is_string($fieldValue)){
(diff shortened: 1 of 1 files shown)


plugin/Filter/FilterBuilderInputVariables.php
```diff
@@ -102,7 +102,9 @@ public function getPostFieldsFilter(array $filterableFields): array
 	private function getFieldFilter($fieldValue, string $type): array
 	{
 		$fieldFilter = [];
-
+        if(is_string($fieldValue)){
```
ogzA, line 105: Wäre es sinnvoll, hier den Null-Status der Variablen abzufragen, oder wäre das überflüssig? z. B. `if($fieldValue !== null && is_string($fieldValue)) `

ogzA (approved):
Ich habe nur eine Frage, ansonsten passt es :)

PR #990, opened 2025-01-27
Bug/P#125533-validator-field-type

mmilawski:
Removes a strict type checking on field names for the validator.
Sometimes it can be that fields have numeric field names instead of strings. In this case it could cause a crash. In some cases it can happen if a field has been deactivated in enterprise.


ingoKrOffice:
sieht gut aus

PR #1012, opened 2025-04-25
Add Release workflow for oo wp-updates

andernath:


andernath:
Must be on master before testing workflow is possible

PR #1039, opened 2025-06-05
bug/#4926465-load-contact-person-not-public

ingoKrOffice:
* fix: load contact person if its broker and not public


mmilawski (approved):
looks good and working as expected.

PR #1173, opened 2026-04-13
perf(P#160717): Verbesserung der API Performance

mmilawski:
This PR improves overall plugin performance by reducing redundant API calls, batching requests more efficiently, and tightening cache behavior. It also includes URL handling fixes for custom ports.

## Measured Improvement (Local)
The following request-count improvements were recorded locally. Results can vary depending on the WP website configuration.

| Page type | Before | After | Reduction | Improvement |
| --- | ---: | ---: | ---: | ---: |
| Estate list | 42 | 17 | 25 | 59.5% |
| Estate detail | 27 | 13 | 14 | 51.9% |
| Generic pages | 7 | 2 | 5 | 71.4% |
| **Total** | **76** | **32** | **44** | **57.9%** |

## Summary
- Introduces per-request in-memory API response caching and request deduplication to reduce duplicate network traffic.
- Optimizes estate/address loading flows by batching queued requests and removing unnecessary follow-up calls.
- Improves cache consistency and robustness for list responses, field dependencies, and pagination metadata handling.
- Fixes detail URL generation to preserve custom ports (for example `:8080`).
- Adds defensive and conditional loading improvements to avoid unnecessary data fetching in forms and list filters.

### API and cache performance
- Added in-memory cache in `SDK/src/internal/ApiCall.php` to reuse identical calls within the same PHP request.
- Deduplicates identical queued requests in a single send cycle and reuses source response/error results.
- Normalizes `fields` requests (including `showfielddependencies`) to improve cache hit consistency.
- Extracted and centralized cached list filtering, sorting, and pagination logic and added safety guards for incomplete payloads.
- Writes cache entries using normalized request parameters for better cache coherence.

### SDK/container/runtime optimizations
- `plugin/SDKWrapper.php` now reuses a shared DI container instead of rebuilding it repeatedly.
- Added in-memory caching of field-type metadata per language.
- `renewCache()` now supports optional language scoping and handles edge cases in `cntabsolute`.
- Introduced helper logic to warm list cache once per list/language when needed (currently not actively used).

### Estate/address request batching
- `plugin/EstateList.php` now batches `idsfromrelation` and `estatepictures` requests before a single `sendRequests()`.
- Estate file handling now processes pre-fetched picture records and avoids redundant request paths.
- `plugin/AddressList.php` now batches formatted/raw address reads and batches count-estate lookups.
- Added optional flag to skip related-estate enrichment when loading broker addresses where not required.

### Conditional field loading and behavior fixes
- Regional supplement field loading is now conditional in `plugin/EstateList.php` and `plugin/Form.php` (only when actually needed/configured).
- `plugin/EstateDetail.php` avoids unnecessary reload work for similar estates and exits early when no estate IDs are present.
- `plugin/Controller/EstateViewSimilarEstates.php` normalizes marketing method values to lowercase for filter matching consistency.

### URL handling
- Added `plugin/Utility/UrlHelper.php` to build base URLs including optional port.
- Updated `plugin/Controller/EstateDetailUrl.php` and `plugin/Controller/AddressDetailUrl.php` to preserve custom ports in generated detail/title URLs.

### DIFF
--- a/SDK/src/internal/ApiCall.php
+++ b/SDK/src/internal/ApiCall.php
@@ -285,8 +285,19 @@ class ApiCall
 	{
 		if ($params["formatoutput"] == true)
 		{
+			if (
+				!isset($cachedResponse["data"]["records"]) ||
+				!is_array($cachedResponse["data"]["records"]) ||
+				!isset($cachedResponse["raw"]["data"]["records"]) ||
+				!is_array($cachedResponse["raw"]["data"]["records"]) ||
+				!isset($cachedResponse["types"]) ||
+				!is_array($cachedResponse["types"])
+			) {
+				return $cachedResponse;
+			}
+
 			$this->filterRecords($cachedResponse, $params["filter"]);
-			if(in_array("sortby", $params) && $params["sortby"] != null)
+			if(array_key_exists("sortby", $params) && $params["sortby"] != null)
 				$cachedResponse["data"]["records"] = $this->sortRecords($cachedResponse, $params["filter"], $params["sortby"], $params["sortorder"] ?? 'ASC');
 			$cachedResponse["data"]["meta"]["cntabsolute"] = count($cachedResponse["data"]["records"]);
 
@@ -449,6 +460,17 @@ class ApiCall
 
 	private function filterRecords(array &$cachedResponse, array $filter)
 	{
+		if (
+			!isset($cachedResponse["data"]["records"]) ||
+			!is_array($cachedResponse["data"]["records"]) ||
+			!isset($cachedResponse["raw"]["data"]["records"]) ||
+			!is_array($cachedResponse["raw"]["data"]["records"]) ||
+			!isset($cachedResponse["types"]) ||
+			!is_array($cachedResponse["types"])
+		) {
+			return;
+		}
+
 		$records = $cachedResponse["data"]["records"];
 		$filteredArray = $records;
 		$filteredArrayRaw = $cachedResponse["raw"]["data"]["records"];
--- a/SDK/src/internal/ApiCall.php
+++ b/SDK/src/internal/ApiCall.php
@@ -37,6 +37,14 @@ class ApiCall
 	/** @var array */
 	private $_curlOptions = array();
 
+	/**
+	 * In-memory cache for API responses within a single PHP request.
+	 * Prevents duplicate identical API calls (e.g. fields, idsfromrelation)
+	 * from hitting the network multiple times per page load.
+	 * @var array<string, array>
+	 */
+	private static $_inMemoryCache = array();
+
 
 	/**
 	 * @param string $actionId
@@ -161,8 +169,24 @@ class ApiCall
 		foreach ($this->_requestQueue as $requestId => $pRequest)
 		{
 			$usedParameters = $pRequest->getApiAction()->getActionParameters();
-			$cachedResponse = $this->getFromCache($usedParameters);
 			$params = $usedParameters["parameters"];
+
+			// 1. Check in-memory cache first (catches ALL duplicate calls within this PHP request)
+			$inMemoryKey = $this->getInMemoryCacheKey($usedParameters);
+			if (isset(self::$_inMemoryCache[$inMemoryKey]))
+			{
+				$inMemoryResponse = self::$_inMemoryCache[$inMemoryKey];
+				// Apply list-specific filtering/sorting if needed (same logic as DB cache hit)
+				if (isset($params['listname']))
+				{
+					$inMemoryResponse = $this->applyListCacheFiltering($inMemoryResponse, $params);
+				}
+				$this->_responses[$requestId] = new Response($pRequest, $inMemoryResponse);
+				continue;
+			}
+
+			// 2. Check DB cache (existing logic for list-based caching)
+			$cachedResponse = $this->getFromCache($usedParameters);
 
 			if ($cachedResponse === null)
 			{
@@ -180,18 +204,12 @@ class ApiCall
 			}
 			else
 			{
+				// Store in in-memory cache for subsequent identical calls
+				self::$_inMemoryCache[$inMemoryKey] = $cachedResponse;
+
 				if($cachedResponse != null && isset($params['listname']))
 				{
-					if($params["formatoutput"] == true)
-					{
-						$this->filterRecords($cachedResponse, $params["filter"]);
-						if(in_array("sortby", $params) && $params["sortby"] != null)
-							$cachedResponse["data"]["records"] = $this->sortRecords($cachedResponse, $params["filter"], $params["sortby"], $params["sortorder"] ?? 'ASC');
-						$cachedResponse["data"]["meta"]["cntabsolute"] = count($cachedResponse["data"]["records"]);
-
-						$cachedResponse["data"]["records"] =
-							$this->recordsPerPage($cachedResponse["data"]["records"], intval($params["listlimit"] ?? 20), intval($params["listoffset"] ?? 0));
-					}
+					$cachedResponse = $this->applyListCacheFiltering($cachedResponse, $params);
 				}
 				$this->_responses[$requestId] = new Response($pRequest, $cachedResponse);
 				$saveToCache = false;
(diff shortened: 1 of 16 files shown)


Nigel991:
Hey Michael, ich hab Copilot in VSCode mal über das diff laufen lassen. Ich glaube die meisten Sachen kannst du ignorieren aber die Punkte 4, 5 und 6 schienen schon wichtig. kannst du mal schauen ob die stimmen?

---

## Critical / High Severity

### 1. Security: HTTP protocol allowed alongside HTTPS (OWASP A02)
In SDKWrapper.php and APIAvailabilityChecker.php:
```php
CURLOPT_PROTOCOLS => CURLPROTO_HTTP | CURLPROTO_HTTPS
```
The original code restricted to HTTPS-only. Now plain HTTP is allowed, meaning API credentials (token/secret) could be sent unencrypted if `ONOFFICE_API_SERVER` is ever changed to `http://` or if DNS is poisoned. This is a downgrade in transport security. If this is only needed for local dev, it should be gated behind a `WP_DEBUG` / environment check, not applied unconditionally in production.

### 2. Static in-memory cache never cleared — unbounded memory growth
`ApiCall::$_inMemoryCache` is `static` and never cleared during a PHP request. For long-running processes (WP-CLI, queue workers, WP-Cron with many tasks) this can grow unboundedly and exhaust memory. Consider adding a `clearInMemoryCache()` method or a size cap.

### 3. Static shared DI container in SDKWrapper is not test-safe
```php
private static $_pSharedContainer = null;
```
This shared container persists across test cases in PHPUnit (which runs in a single process). If any test mutates the container or its services, all subsequent tests are affected. Similarly, `EstateList::getGlobalContainer()` has the same issue. There's no `reset()` method exposed for testing.

---

## Medium Severity

### 4. `$params["parameters"]` may not exist — potential `Undefined index`
In `collectOrGatherRequests`, after `normalizeFieldDependencyParameters`:
```php
$params = $usedParameters["parameters"];
```
If the request has no `parameters` key (e.g., a non-fields request), this will produce a PHP notice/warning. The original code called `getActionParameters()` which always returns the full array, but `normalizeFieldDependencyParameters` only sets `parameters` for `fields` resource types.

### 5. `applyListCacheFiltering` accesses `$params["filter"]` without checking it exists
If a list request has `formatoutput == true` but no `filter` key, this will error:
```php
$this->filterRecords($cachedResponse, $params["filter"]);
```

### 6. In-memory cache stores response data that may be mutated in-place
`filterRecords` takes `$cachedResponse` by reference (`array &$cachedResponse`). When a response is stored in the in-memory cache and then later retrieved, `applyListCacheFiltering` calls `filterRecords()` which **mutates** the passed array. Since the in-memory cache stores the same array reference, the cached copy gets corrupted for subsequent cache hits. You likely need to deep-clone (e.g., `$inMemoryResponse = unserialize(serialize(self::$_inMemoryCache[$inMemoryKey]))`) before applying filtering.

### 7. Duplicate deduplication has a subtle ordering issue
Duplicate requests for the same cache key are resolved from the source request's response. But if the source request *fails* and lands in `$this->_errors`, the duplicates get the error propagated — which is correct. However, if the source request **neither** succeeds nor fails (e.g., the HTTP response is missing that particular request index due to a partial API failure), duplicates silently get **no response and no error**, leading to missing data with no indication of failure.

### 8. `$saveToCache` variable removed from DB-cache branch
The original code set `$saveToCache = false` when a DB cache hit was found, preventing the response from being re-written to cache. The new code removed this line. Now `sendHttpRequests` is called with the original `$saveToCache` value, which is fine for HTTP responses. But the in-memory cache storage at the end also stores **all** HTTP responses unconditionally when `$saveToCache` is true, regardless of whether they came from the DB cache path — this is probably intentional but worth confirming.

### 9. `normalizeFieldDependencyParameters` applied to `createRequest()` output
```php
$parametersThisAction = $this->normalizeFieldDependencyParameters(
    $pRequest->createRequest($token, $secret)
);
```
`createRequest()` returns the signed request payload (with HMAC, timestamp, etc.). Applying `normalizeFieldDependencyParameters` to this **mutates the signed payload** by adding `showfielddependencies=true` and re-sorting parameters. If the API validates the HMAC against the parameters, this would break authentication. Verify that `createRequest()` returns parameters **before** HMAC signing, or that the HMAC doesn't cover the parameters array.

---

## Low Severity / Code Quality

### 10. `EstateDetail::getSimilarEstates()` — `loadEstates()` removed but `resetEstateIterator()` used
```php
$pCopyThis = clone $this;
$pCopyThis->resetEstateIterator();
```
The old code called `setFormatOutput(false)` then `loadEstates()`. The new code just resets the iterator. This assumes the estate data is already loaded, which should be the case since `getSimilarEstates()` is called during rendering. But the early return `if ($this->getEstateIds() === [])` guards this correctly.

### 11. `addFieldsAddressEstateWithRegionValues` selectively loaded — potential regression
In EstateList.php, the call to `addFieldsAddressEstateWithRegionValues` was removed from two methods and made conditional in a third (only if `regionaler_zusatz` is in filterable fields). Similarly in Form.php, it's conditional on the form using the field. If any template or filter references `regionaler_zusatz` indirectly (e.g., via a computed field or dependency), the field definition won't be available, causing silent data loss. Ensure test coverage exists for views/forms using this field.

### 12. `EstateViewSimilarEstates` — marketing method lowercased
```php
$marketingMethod = (string)($pValuesCurrentEstate->getValueRaw('vermarktungsart') ?? '');
$this->_pFilterConfiguration->setMarketingMethod(mb_strtolower($marketingMethod));
```
This is a fix (standardizing case), but if the API filter is case-sensitive and expects the original casing, this could break similar estate filtering. Verify the API treats `vermarktungsart` case-insensitively.

### 13. `renewCache` language parameter — no validation
The new `$languages` parameter accepts arbitrary strings with no validation against supported languages. Passing invalid language codes would trigger API calls that return empty/error results and cache them.

### 14. `cntabsolute` defensive handling is odd
```php
$cntAbsolute = $resultMeta['cntabsolute'] ?? 0;
if (is_array($cntAbsolute)) {
    $cntAbsolute = $cntAbsolute[0] ?? 0;
}
```
If `cntabsolute` can be an array, that suggests an upstream API inconsistency. This silently picks the first element. A log warning would help diagnose when this unexpected shape occurs.

### 15. `in_array` changed to `array_key_exists` for sort check
Original: `in_array("sortby", $params)` → New: `array_key_exists("sortby", $params)`. This is a **correct fix** — `in_array` checks *values*, not keys. Good catch.

### 16. `UrlHelper::buildBaseUrl` — no validation of array keys
```php
$baseUrl = $urlElements['scheme'] . '://' . $urlElements['host'];
```
If `wp_parse_url` returns `false` or is missing `scheme`/`host`, this will produce PHP warnings. While `wp_parse_url` should always return valid parts for well-formed URLs, defensive checks would be safer.

---

## Summary

| Severity | Count | Key Issues |
|----------|-------|------------|
| **Critical** | 1 | HTTP protocol allowed for API calls (security downgrade) |
| **High** | 2 | Unbounded static cache growth; static container breaks test isolation |
| **Medium** | 6 | Cache mutation via reference, undefined index risks, HMAC integrity concern |
| **Low** | 7 | Conditional field loading regressions, minor robustness issues |

The most actionable item is **#6 (cache mutation by reference)** — this is a functional bug that will cause incorrect list filtering results when the same list data is requested multiple times with different filter/sort/pagination parameters within one PHP request. 

Completed: *Compile findings report* (7/7)

mmilawski:
Bin die einzelnen Sachen durchgegangen, paar Sachen waren schon valid, habe ich gefixt:

### Scope
- **1 (HTTP/HTTPS)** intentionally ignored for now (needed for local proxy, not a problem 

### Point-by-point status
| # | Verdict | Status | Notes |
|---|---|---|---|
| 2 | Partly valid (long-running processes) | Not changed | `ApiCall::$_inMemoryCache` is static; for normal WP requests this is usually request-bounded, but CLI/long workers could still grow. We are currently not using long running processes on any wp website |
| 2 | Not a problem | Not changed | |
| 4 | Not valid | No change | `parameters` exists by design via `ApiAction` (`SDK/src/internal/ApiAction.php`). |
| 5 | Valid | ✅ Fixed | Added safe fallback for missing `filter` in `applyListCacheFiltering` (`SDK/src/internal/ApiCall.php`). |
| 6 | Not valid in current implementation | No change | In-memory cache value is copied to local var before filtering, and `applyListCacheFiltering` takes/returns arrays by value (no direct shared-reference mutation issue in this path). |
| 7 | Valid | ✅ Fixed | Added handling for partial/misaligned API result arrays: missing queued actions now get a synthetic error instead of silently disappearing (`SDK/src/internal/ApiCall.php`). |
| 8 | Informational | No change | Current cache-write behavior is intentional after previous fixes. |
| 9 | Not valid | No change | HMAC generation uses timestamp/token/resourcetype/actionid only (`SDK/src/internal/Request.php`), not request `parameters`. |
| 10 | Low risk | No change | `getSimilarEstates()` has empty-estate guard and uses existing loaded state; current behavior is acceptable. |
| 11 | Potential regression risk | ✅ Fixed defensively | Reintroduced conditional region metadata loading where `regionaler_zusatz` is actually needed in estate field/filter pipelines and broadened form check to include inputs/required/hidden fields (`plugin/EstateList.php`, `plugin/Form.php`). |
| 12 | Unclear | No change | Lowercasing `vermarktungsart` kept as-is (works with current raw values and filter behavior). |
| 13 | Valid (robustness) | ✅ Fixed | Added strict language normalization/validation in cache prewarm path; invalid language codes are filtered and fallback to default language (`plugin/SDKWrapper.php`). |
| 14 | Low severity | No change | `cntabsolute` array handling remains defensive. |
| 15 | Correct fix | Already done | `array_key_exists` is used for `sortby` key check. |
| 16 | Valid (robustness) | ✅ Fixed | `UrlHelper` now validates parsed URL parts; controllers use safe path/base extraction and graceful fallback when URL parts are missing (`plugin/Utility/UrlHelper.php`, `plugin/Controller/EstateDetailUrl.php`, `plugin/Controller/AddressDetailUrl.php`). |
### Summary of fixes actually applied from the review
- **5**, **7**, **11**, **13**, **16** addressed.
- **1** intentionally not applied for now.
- Remaining open items are either low-risk, informational, or not reproducible as concrete bugs in the current code.

Nigel991 (approved):
Sieht gut aus 👍

PR #1242, opened 2026-07-09
fix(P#168033): Geo Filter

mmilawski:
The geo filter was completely broken: it never actually filtered results by distance and always returned the full, unfiltered result set regardless of the search radius.

### Root causes

- **Inverted coordinates in distance calculation:** `Coordinate(latitude, longitude)` was called with `(longitude, latitude)`, causing all distance calculations to be wildly off.
- **Result set too small for client-side filtering:** The API only returned the default page size (~20 records). Geo filtering is applied client-side on the full result set, so it had too few records to produce meaningful results.
- **Cache key collisions:** The cache key only included the `Id` filter, so changing geo or other filters would reuse a stale cached response.
- **Missing raw/types data blocks:** `filterRecords` assumed `raw.data.records` and `types` were always present, but geo-filtered responses from the cache don"t include them.

### Changes

**SDK - Coordinate order (`SDK/src/internal/ApiCall.php`)**
- Changed `Coordinate(lng, lat)` to `Coordinate(lat, lng)` - the first argument to `Coordinate` is latitude, not longitude.

**SDK - Null-safe and auto-generated data (`SDK/src/internal/ApiCall.php`)**
- Added null-coalescing for `min`/`max` keys in geo filter arrays.
- `filterRecords` now auto-generates `raw.data.records` and `types` when missing from the response.

**SDK - Fresh response geo filtering (`SDK/src/internal/ApiCall.php`)**
- New `applyFreshGeoFiltering` method applies client-side geo filtering to fresh responses (not just cached ones) after they are written to cache.

**SDK - Geo bypass for structure validation (`SDK/src/internal/ApiCall.php`)**
- `applyListCacheFiltering` skips `raw.data.records`/`types` validation when a geo filter is present, since those structures are only needed for non-geo filtering.

**Cache - All filters in cache key (`plugin/Cache/DBCache.php`)**
- Cache key now includes all filter parameters (serialized), not just `Id`. Prevents stale cache hits when filters change.

**List views - Full result sets (`plugin/EstateList.php`, `plugin/AddressList.php`)**
- When `geo_search` is present, `numRecordsPerPage` is increased to 500 to fetch the maximum possible records for client-side distance filtering.
- AddressList additionally resets the offset and applies pagination slicing after filtering.

### Tests

- `SDK/tests/ApiCallTest.php`: 5 new tests covering coordinate order, missing raw/types generation, null-safe min/max, geo bypass logic, and geo distance filtering.
- `tests/TestClassDBCache.php`: New file with 5 tests covering cache key determinism, filter-aware keys, and geo filter inclusion.

### DIFF
--- a/SDK/src/internal/ApiCall.php
+++ b/SDK/src/internal/ApiCall.php
@@ -373,7 +373,6 @@ class ApiCall
 
 	/**
 	 * Apply client-side geo filtering to fresh list responses after cache writing.
-	 * Fresh API responses are already paginated by the API, so keep that slice.
 	 *
 	 * @param Request[] $actionParametersOrder
 	 */
@@ -403,7 +402,7 @@ class ApiCall
 
 			$pResponse = $this->_responses[$requestId];
 			$params['listoffset'] = 0;
-			$params['listlimit'] = PHP_INT_MAX;
+			$params['listlimit'] = 500;
 			$this->_responses[$requestId] = new Response(
 				$pRequest,
 				$this->applyListCacheFiltering($pResponse->getResponseData(), $params)
--- a/SDK/src/internal/ApiCall.php
+++ b/SDK/src/internal/ApiCall.php
@@ -839,6 +839,7 @@ class ApiCall
 			$cachedResponse["data"]["records"] = $filteredArray;
 			$cachedResponse["raw"]["data"]["records"] = $filteredArrayRaw;
 			$filteredArray = $this->sortRecords($cachedResponse, $filter, 'geo_distance', 'ASC');
+			$filteredArray = array_slice($filteredArray, 0, $isGeoAndMax);
 		}
 		$cachedResponse["data"]["records"] = $filteredArray;
 		$cachedResponse["raw"]["data"]["records"] = $filteredArrayRaw;
--- a/SDK/src/internal/ApiCall.php
+++ b/SDK/src/internal/ApiCall.php
@@ -121,6 +121,7 @@ class ApiCall
 		{
 			return;
 		}
+
 		foreach ($actionParameters as $key => $action) {
 			if (isset($action['parameters']['filter']['geo'])) {
 				unset($actionParameters[$key]['parameters']['filter']['geo']);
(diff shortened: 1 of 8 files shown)


Nigel991 (dismissed):
Von meiner Seite aus sieht es gut aus. Die KI hat noch ein paar Sachen gefunden die aber alle  ignoriert werden können. Nur einen Punkt der KI hab ich nicht ganz verstanden, vielleicht kannst du es dir nochmal durchlesen vor dem Merge:

## Geo search: guard mismatch blanks page 2+ when `geo_search` is empty/malformed

### Severity
Medium

### Affected files
- `wp-content/plugins/oo-wp-plugin/plugin/EstateList.php`
- `wp-content/plugins/oo-wp-plugin/plugin/AddressList.php`

### Summary
The geo-radius search uses two different conditions to guard the same feature:

- **Request build** switches to "fetch 500 / offset 0" mode only on a *validated* geo
  search — `$hasGeoSearch`, which requires `count($geoCoords) === 2`.
- **Client-side re-slice** runs on the *looser* `isset($_GET['geo_search'])`.

When `geo_search` is present but not exactly two comma-separated parts, the two guards
disagree.

### Locations
| Concern | EstateList.php | AddressList.php |
|---|---|---|
| Request build (`$hasGeoSearch`) | L834–865 | L269–280 |
| Client-side slice (`isset($_GET['geo_search'])`) | L376 | L377 |

### Steps to reproduce
1. Run a geo search, then clear the geo field but leave an empty `?geo_search=` in the URL
   (WP pagination links typically preserve query args).
2. Navigate to page 2.

### What happens
- `explode(',', '')` → `['']`, so `count !== 2` → `$hasGeoSearch = false`.
- Request build falls back to normal server-side pagination: page 2 returns its own
  ~20 records as array indices `0..19`.
- The slice still fires (`isset` is true) and computes
  `array_slice($records, ($page - 1) * $perPage, $perPage)` → offset `20` on a 20-element
  array → **empty result page**.

### Expected
An empty/malformed `geo_search` should behave exactly like no geo search: normal
pagination on every page.

### Suggested fix
Gate the slice on the same validated condition as the request build instead of bare
`isset()` — either recompute the 2-coordinate check at the slice site, or thread the
validated `$hasGeoSearch` flag through to it so both paths agree.

Nigel991 (dismissed):
👍

PR #1308, opened 2026-09-02
fix: clear the cache on demand and cache searched estate lists

mmilawski:
## TL;DR

Two independent cache defects, both reproduced on live boxes:

1. **"Clear onOffice cache" does not clear the cache.** It only schedules a cron job. On hosts that set `DISABLE_WP_CRON` and run wp-cron from the system crontab, nothing happens for up to 10 minutes — while the success notice appears immediately. The button now clears synchronously (37 ms measured), so it behaves the same on every host.
2. **Estate lists with an active search are never cached.** Every single page view hits the onOffice API live, forever. They are now served from the entry the cron already warmed (135–169 ms → 0–5 ms measured, identical results).

Plus a smaller one: `clearAll()` left behind every entry written in the same second as the click.

6 new tests, 1348 green.

---

## 1. The clear-cache button never cleared anything

`plugin.php` did not clear the cache. It unscheduled and rescheduled `oo_cache_renew`, set the "cache was cleaned" notice, and redirected:

```php
wp_schedule_event(time(), $onofficeSettingsCache, 'oo_cache_renew');
update_option('onoffice-notice-cache-was-cleared', true);
```

Whether anything actually happened then depends entirely on how the host runs WP-Cron. Measured on three environments:

| Environment | `DISABLE_WP_CRON` | wp-cron runner | Delay after the click |
|---|---|---|---|
| Raidboxes A | `true` | system crontab `*/10` | **0–10 minutes** |
| Raidboxes B | `false` | page load | ~1 second |
| IONOS | not set | supervisor loop, 60 s | ~1–60 seconds |

Same plugin version, same site content — only the host configuration differs. This is why the button is reported as broken on some sites and working on others, and why the report is hard to reproduce.

Verified directly on the slow box: `rows=125` before the click, `rows=125` immediately after, unchanged.

**Fix:** call `CacheHandler::clear()` in the request. It is one `DELETE` plus a handful of `delete_transient()` calls — 37 ms measured, so the redirect is not noticeably delayed and the success notice is now truthful. The warm-up stays out of band via `wp_schedule_single_event()`.

The old code also unscheduled the recurring event *before* rescheduling it. `wp_schedule_event()` returns `false` for a recurrence name that no `cron_schedules` entry provides, so a failed reschedule left the site with no cache renewal at all — permanently, since the reschedule is only attempted when the event is missing or the setting changes. The recurring event is no longer touched.

## 2. `clearAll()` did not clear all

```php
$this->_options['ttl'] = 0;
$this->cleanup();       // DELETE ... WHERE UNIX_TIMESTAMP(cache_created) < %d
```

With a TTL of 0 the cutoff is `time()`, and `<` does not match rows written in the current second. Reproduced: write two entries, call `clearAll()`, both survive.

The intent was always "delete everything" — `88aa3fb0` ("fix bug clear cache") already had to add a `cleanCache` flag to stop the cache-duration setting from overriding the zero TTL. The boundary was missed. `clearAll()` is now a plain unconditional `DELETE`, which also makes the `cleanCache` flag unnecessary.

`cleanup()` — the method where the TTL is meaningful — is unchanged.

## 3. Searched estate lists were never cached

The DB cache key for a list request is `md5(listname | formatoutput | outputlanguage | filter)`.

- The cron warms the cache using `DefaultFilterBuilder::getDefaultFilter()` — the list's default filter.
- The frontend reads using `DefaultFilterBuilder::buildFilter()` — the default filter **plus** the active search criteria, geo filter, favourites ids, and so on.

As soon as a visitor searches, the filter differs, so the key differs and the lookup misses. The response is fetched live, and `writeCacheForResponses()` then refuses to store it because it is a truncated page:

```php
if ($isBeyondFirstPage || $isTruncated) {
    continue;
}
```

So the entry is never created, and the next visitor repeats the whole thing. Measured on a test box after warming the cache — the searched list is consistently slower and produces no list cache entry:

```
GET /immobilien/                      0.81s  0.83s
GET /immobilien/?vermarktungsart=kauf 1.16s  1.14s
```

**Fix:** a list request already carries the parameters the cron warmed it under, in `params_list_cache`. The cache is now looked up under those, and `applyListCacheFiltering()` trims the full set down to the requested page — exactly what that method already does on the unsearched path.

### Why there are two keys

```php
$cacheKey   = $this->getInMemoryCacheKey($cacheParameters);
$requestKey = $this->getInMemoryCacheKey($usedParameters);
```

`$cacheKey` is deliberately shared by every page and every search of one list — that sharing is the point. `$requestKey` stays unique per request and is the only key `$sourceRequestIdByCacheKey` may use.

This matters: duplicate detection keyed by `$cacheKey` makes two different searches on the same list look like duplicates of each other, and the second one is served the first one's records — wrong estates, not just stale ones. `TestClassApiCall::testDifferentSearchesOnTheSameListAreNotTreatedAsDuplicates` fails with a single key and passes with two.

Requests without `params_list_cache` produce identical keys, so everything outside list views behaves exactly as before.

## Verification

Every search combination was compared against a cache-free live API call on a real box — identical estate ids in all cases:

| Search | live ids | cached ids | live | cached |
|---|---|---|---|---|
| (no search) | 16 | 16 | 153 ms | 5 ms |
| `vermarktungsart=kauf` | 16 | 16 | 169 ms | 0 ms |
| `vermarktungsart=miete` | 2 | 2 | 135 ms | 0 ms |
| `objektart in haus,wohnung` | 16 | 16 | 137 ms | 0 ms |
| `kaufpreis 100k..400k` | 6 | 6 | 135 ms | 0 ms |
| `anzahl_zimmer >= 3` | 11 | 11 | 146 ms | 0 ms |

The three raw-output paths were checked and none of them reach the new lookup, so the `formatoutput` guard inside `applyListCacheFiltering()` cannot be bypassed:

- `EstateList::loadEstates()` unsets `params_list_cache` for the raw request
- `fetchDataForOrderEstatesByTags()` requires `$formatOutput === true`
- `getEstateParametersForMap()` never sets `params_list_cache`

## Tests

- `tests/TestClassDBCache.php` — `clearAll()` removes entries written in the same second, and does so regardless of the cache-duration setting
- `tests/TestClassApiCall.php` — different searches on one list stay separate; the same search still deduplicates; requests without `params_list_cache` behave unchanged

1348 tests green (1342 before).

Note: `SDK/tests/ApiCallTest.php` is not covered by `phpunit.xml.dist` — it is neither under `./tests/` nor uses the `Test` prefix — so it never runs in CI. The new SDK tests were placed in `tests/` to actually be part of the gate.

## Risk

Item 3 touches the cache lookup that every request on every site goes through. The safety property is that both keys are identical for anything without `params_list_cache`, which is every request outside estate list views.

### DIFF
--- a/SDK/src/internal/ApiCall.php
+++ b/SDK/src/internal/ApiCall.php
@@ -225,11 +225,27 @@ class ApiCall
 			);
 			$params = $usedParameters["parameters"];
 
-			// 1. Check in-memory cache first (catches ALL duplicate calls within this PHP request)
-			$inMemoryKey = $this->getInMemoryCacheKey($usedParameters);
-			if (isset(self::$_inMemoryCache[$inMemoryKey]))
+			// params_list_cache holds the parameters the cron warmed the list under:
+			// same list, default filter, complete record set. Look the cache up under
+			// those, so a request with an active search hits the entry at all, and let
+			// applyListCacheFiltering() below trim the full set to this page.
+			$cacheParameters = $usedParameters;
+			if (isset($params['params_list_cache']) && is_array($params['params_list_cache']))
 			{
-				$inMemoryResponse = self::$_inMemoryCache[$inMemoryKey];
+				$cacheParameters['parameters'] = $params['params_list_cache'];
+			}
+
+			// $cacheKey is shared by every page and search of one list, $requestKey is not.
+			// Duplicate detection must use $requestKey: keyed by $cacheKey, two different
+			// searches on the same list look like duplicates and the second one gets the
+			// first one's records. See TestClassApiCall.
+			$cacheKey = $this->getInMemoryCacheKey($cacheParameters);
+			$requestKey = $this->getInMemoryCacheKey($usedParameters);
+
+			// 1. Check in-memory cache first (catches ALL duplicate calls within this PHP request)
+			if (isset(self::$_inMemoryCache[$cacheKey]))
+			{
+				$inMemoryResponse = self::$_inMemoryCache[$cacheKey];
 				// Apply list-specific filtering/sorting if needed (same logic as DB cache hit)
 				if (isset($params['listname']))
 				{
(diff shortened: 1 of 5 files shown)


ingoKrOffice (approved):
Sieht gut aus

